PR #29, opened 2020-10-13
GHO-11: add the source link

rupl:
# GHO-11

I spent an ungodly amount of time trying to format the link but I failed so here's a PR with the text hyperlinked.

RTL layout also included in the PR now. You have to create an Arabic translation of the page to see the effect, because otherwise the `<article>` for the Article content type will itself have `dir="ltr"` as an attribute, which applies both sets in a rather strange way. I'm not sure if we can/should safeguard against that situation. It seems they'll have translations for everything.


orakili (approved):
Looks good. I will add a field formatter to display the field data in accordance to the prototype in another PR.

PR #30, opened 2020-10-13
GHO-39: add pre-title using a method from W3 recommendations

rupl:
# GHO-39

The new method we settled on for pre-title, per ticket discussion. I think we only had two of these, so the refactor was pretty minor.


orakili (approved):
That's great, thanks.

PR #31, opened 2020-10-14
[GHO-11] Dataset link formatter

orakili:
Ticket: GHO-11

This adds a "gho dataset link" field widget and formatter. The widget is just to be able to have a better label for the link ("Name" instead of "Link text") and the formatter (with its template), is basically to allow us to distinguish between dataset links and other links.

This was necessary because there was no way to override the link template (`link-formatter-link-separate.html.twig`) only for the dataset links (meaning, the override would have been applied to any link fields using the `link separate` formatter).


html/modules/custom/gho_fields/templates/gho-dataset-link-formatter.html.twig
```diff
@@ -0,0 +1,18 @@
...
+ * - title: Source name.
+ * - url: Dataset url.
+ *
+ * @ingroup themeable
+ */
+#}
+{% set link = link('Download dataset'|t, url, {})|render %}
+<div class="gho-dataset-link">
+{% trans %}
+  Source: {{ source }}. {{ link }}.
```
rupl, line 16: so simple once all the other stuff is installed 😆

rupl (approved):
*chef's kiss*

html/modules/custom/gho_fields/templates/gho-dataset-link-formatter.html.twig
rupl: Whoops, I got too excited. The word "Source" is no longer translatable. I can follow-up.

rupl: ahhh I see now.. the `{% trans %}` tag takes care of it

orakili: @rupl I wrapped everything in a `%trans%` so that the entire sentence `Source: {{ source }}. {{ link }}.` can be translated. I thought that would better than translating each component separately because the order of the words, the spaces and punctuations would probably differ from one language to the other (ex: in French: `Source : OCHA. Téléchargez les données.` <- note the space surrounding the `:`).

Ref: https://getlevelten.com/blog/mark-carver/drupal-8-twig-templates-and-translations

rupl: Yes that all makes sense. I am familiar with the French convention for colons and bangs, since HID API does it that way.

and wow, I haven't seen that URL in a long time.. I worked at L10 before they got into Drupal. my first job out of Uni!

PR #32, opened 2020-10-15
GHO-8: Separator paragraph type

rupl:
# GHO-8

Horizontal separator, which can be manually placed as a Paragraph within the content flow.

I wrote some notes in the ticket but ultimately felt like it was too complicated and fragile to try and inject separators automatically.


orakili (approved):
That works well, thanks.

Just for info, an alternative to using a field for the size could have been to use different view modes and enable the "Paragraphs view mode" behavior in /admin/structure/paragraphs_type/separator

PR #33, opened 2020-10-16
GHO-14: Related Articles

rupl:
# GHO-14

Allows an editor to add entity references that appear as Related Articles. I chose a standard field as it felt most sensible, and the examples in the PDF always appear at the end of the Article.

Excuse the use of weird floats. It would be grand to use CSS Grid but IE doesn't support it well. Combined with the fact that the title is first in source order, I had to get a little creative. Better than untwisting how to move things around in twig, haha.

## Testing

```
$ drush cim
$ drush cr # for new theme library
```

After it's set up you have to add some Article references to an Article. I tried to be thorough; testing in LTR and RTL on Chrome, FF, and IE11. Mobile tested in Chrome devtools and real iOS Safari.

I attached a screencast of the the responsive image styles in action **to the ticket**, and here is the patch I used to test if you wish to try it yourself.

```diff
diff --git html/themes/custom/common_design_subtheme/components/gho-related-article/gho-related-article.css html/themes/custom/common_design_subtheme/components/gho-related-article/gho-related-article.css
index d02eabe..f94ffc5 100644
--- html/themes/custom/common_design_subtheme/components/gho-related-article/gho-related-article.css
+++ html/themes/custom/common_design_subtheme/components/gho-related-article/gho-related-article.css
@@ -108,3 +108,21 @@
     /*margin-right: .333em;*/
   }
 }
+
+.width-tester {
+  color: white;
+  background: red;
+  width: calc(100vw - 24px);
+}
+@media screen and (min-width: 768px) {
+  .width-tester {
+    background: green;
+    width: calc(((100vw - 48px) * .4) - 1rem);
+  }
+}
+@media screen and (min-width: 1008px) {
+  .width-tester {
+    background: blue;
+    width: calc(((960px) * .4) - 1rem);
+  }
+}
diff --git html/themes/custom/common_design_subtheme/templates/nodes/node--article--related-article.html.twig html/themes/custom/common_design_subtheme/templates/nodes/node--article--related-article.html.twig
index 5fc0c73..d3458d0 100644
--- html/themes/custom/common_design_subtheme/templates/nodes/node--article--related-article.html.twig
+++ html/themes/custom/common_design_subtheme/templates/nodes/node--article--related-article.html.twig
@@ -84,6 +84,7 @@
   ]
 %}
 <article{{ attributes.addClass(classes) }}>
+  <div class="width-tester">testing img width</div>
   <h3{{ title_attributes.addClass('gho-related-article__title') }}>
     <a href="{{ url }}" rel="bookmark">{{ label }}</a>
   </h3>
```

### DIFF
--- a/config/core.entity_form_display.node.article.default.yml
+++ b/config/core.entity_form_display.node.article.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.node.article.field_author
     - field.field.node.article.field_hero_image
     - field.field.node.article.field_paragraphs
+    - field.field.node.article.field_related_articles
     - field.field.node.article.field_summary
     - field.field.node.article.field_thumbnail_image
     - node.type.article
@@ -48,10 +49,20 @@ content:
     weight: 7
     settings:
       preview_view_mode: default
-      nesting_depth: '4'
+      nesting_depth: 4
       require_layouts: 0
     third_party_settings: {  }
     type: layout_paragraphs
+    region: content
+  field_related_articles:
+    weight: 26
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    type: entity_reference_autocomplete
     region: content
   field_summary:
     weight: 6
--- a/config/core.entity_view_display.media.image.related_article_thumb.yml
+++ b/config/core.entity_view_display.media.image.related_article_thumb.yml
@@ -1,0 +1,38 @@
+uuid: 4e2f4967-1ce3-4140-8102-797d08065d52
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.media.related_article_thumb
+    - field.field.media.image.field_credits
+    - field.field.media.image.field_media_image
+    - media.type.image
+    - responsive_image.styles.half_width
+  module:
+    - layout_builder
+    - responsive_image
+third_party_settings:
+  layout_builder:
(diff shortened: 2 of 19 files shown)


orakili (approved):
Looks good. A few things that could be changed/added in a subsequent PR:

1. I think the label of the field_related_articles should be set to "visually-hidden" and set to a h2 in `field--node--field-related-articles.html.twig`. Otherwise there is gap in the document hierarchy and it's not clear what those articles are.
2. Nitpicking a bit: the the related article view mode for the media is named `Related Article Thumb` while the associated responsive image style is `Related Article Thumbnail`.
3. Question: Why the 960px max width? 

In the PDF, it says those related articles use the same design as the "at glance" section which seems to use the full width  (see the full width separator between them):

Screen Shot 2020-10-20 at 16 13 06

rupl:
1. Good idea, since there's no title at the moment 😳  but I also see that the title sometimes displays (see p15 of PDF v2).
2. We can rename, no problem: 07c8d80
3. I thought that I'd seen it in the PDF as less than content column width, but now double checking I see it's the other content that is a bit skinnier. Whoops. However I did like the max-width because it contains the text to the shape implied in the PDF. Maybe we can ask Edgar in a few minutes.
4. I totally missed the separator. Nice catch. Also a follow-up because I tried to drop it in, then noticed the images aren't precisely aligned. It becomes noticeable with the separator.

rupl:
I ended up fixing the spacing and adding the separator. Other issues added to the smart checklist of the ticket.

PR #34, opened 2020-10-16
GHO-13: Further reading

rupl:
# GHO-13

New content type, paragraph type, and theming for Further reading list.

I have to say upfront: I am not _in love_ with the solution here, but I wanted to create a list of links which could be moved around as one unit, while also allowing the list to be reordered within the larger component. Ensuring a simple-ish form while satisfying those two UX issues is what motivated me to use the following solution.

I tried a couple methods:

- Initially I made a very basic Paragraph type containing a Link and a Label. Without removing the `.field__item` wrapper on all Paragraphs within Article, I couldn't easily style the successive links. Plus, it was possible to move one link without the others being taken for the ride.
- I tried using the `double_field` plus the formatter that we went with on Bottom figures.. the UX was so ugly when entering content, very unclear to me even as the creator of the widget
- **Solution in PR:** I settled for a list of entity references, using the mandatory Title field as the "Author / Org" credit line. This lets them be created on the fly, translated easily, and be re-used as an existing entity on multiple pages.

image

## Testing

```sh
$ drush cim
$ drush cr # for theme library
```

Tested LTR/RTL on macOS Chome/FF, Win7 IE11.

### DIFF
--- a/config/core.base_field_override.node.further_reading.promote.yml
+++ b/config/core.base_field_override.node.further_reading.promote.yml
@@ -1,0 +1,22 @@
+uuid: ec3a030b-6da7-4e59-92db-beae2f571a51
+langcode: en
+status: true
+dependencies:
+  config:
+    - node.type.further_reading
+id: node.further_reading.promote
+field_name: promote
+entity_type: node
+bundle: further_reading
+label: 'Promoted to front page'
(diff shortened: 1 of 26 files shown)


rupl:
Closing in favor of #37

PR #35, opened 2020-10-19
GHO-12: Appeals tags

rupl:
# GHO-12

PR contains config/theming for Appeals tags. They're attached to Article nodes but I think the template names I chose to overrode mean it will work anywhere (as paragraph, on different content types, etc)

The colors are stored in the component-specific CSS and they came straight off of hum-insight.info production. 

## Testing

```sh
$ drush cim
$ drush cr
```

After importing config, create taxonomy terms with the following titles: HRP, FA, RRP, Other — they should automatically pick up the appropriate colors and appear like so:

image

I used similar styles to some of the other flex components (negative margin on container) to ensure that RTL works out of the box. The tags themselves are English only and not translatable. I used the [Arabic-language 2020 GHO PDF](https://www.humanitarianresponse.info/sites/www.humanitarianresponse.info/files/documents/files/gho-2020_ar_web.pdf) (see page 28) as the basis for my decision, which uses latin-character tags amongst the otherwise arabic content.

### DIFF
--- a/html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
+++ b/html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
@@ -31,6 +31,6 @@
   ]
 %}
 {% set tagname = name|render|striptags|trim %}
-<div{{ attributes.setAttribute('id', 'taxonomy-term-' ~ term.id).addClass(classes, 'gho-appeals-tag--' ~ tagname|lower) }} title="{{ content.field_appeal_extended|render|striptags|trim }}">
+<div{{ attributes.addClass(classes, 'gho-appeals-tag--' ~ tagname|lower) }} title="{{ content.field_appeal_extended|render|striptags|trim }}">
   {{- tagname -}}
 </div>
--- a/html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
+++ b/html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
@@ -31,6 +31,6 @@
   ]
 %}
 {% set tagname = name|render|striptags|trim %}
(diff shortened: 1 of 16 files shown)


rupl:
After taking a second look at the Arabic PDF linked above, I've allowed translation on the extended name field. The PDF lists translations at the bottom of page 28 but since it's no trouble to do that inline, we can offer title attributes once the tag gets translated. We will instruct folks not to alter the main tag name (e.g. `HRP`)

Here's that it looks like on mouse hover:

Screen Shot 2020-10-19 at 16 48 18

html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
```diff
@@ -0,0 +1,36 @@
...
+#}
+{%
+  set classes = [
+    'taxonomy-term',
+    'vocabulary-' ~ term.bundle|clean_class,
+    'gho-appeals-tag',
+  ]
+%}
+{% set tagname = name|render|striptags|trim %}
+<div{{ attributes.addClass(classes, 'gho-appeals-tag--' ~ tagname|lower) }} title="{{ content.field_appeal_extended|render|striptags|trim }}">
```
orakili, line 34: Rather than `tagname|lower`, better to use `tagname|clean_class` which changes the text to lower but also ensures it's suitable to use as a class.

Also are the `|render|striptags` necessary on the `name` necessary?

orakili (changes requested):
Looks good, just a tiny change to make in the template (see comment).

html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
rupl: nice tip. I have seen `clean_class` but didn't think to use it.

yes on the render and strip tags. At least in the sense that it's the only combo I could find that totally removes all whitespace to make the value suitable as a `title` attribute.

rupl:
Added the fifth color per our meeting/demo with Edgar today:

image

orakili (approved):
Looks good. Thanks!

PR #36, opened 2020-10-21
[GHO-15] Footnotes

orakili:
Ticket: GHO-15

This adds footnotes to text paragraphs. The current approach is to bundle a text paragraph with its footnotes.

Screen Shot 2020-10-21 at 17 50 54

There a few parts to make that work:

**Footnotes field**

- A `footnotes` field on the text paragraph type which uses ckeditor with a very restricted options defined in a `footnotes` text format. This allows simple formatting: bold, italic and links.
- The `ckeditor_config` module has been added so that we can add a class to the body inside the ckeditor iframe to styles the footnotes. 

The trick here, is to use a CSS counter to add a number to the paragraphs in the ckeditor of the footnotes field to give the impression it's a list:

Screen Shot 2020-10-21 at 17 38 59

Editors can then reference the footnotes by adding a simple `[X]` tag where appropriate in the text field:

Screen Shot 2020-10-21 at 17 40 17

The text and the footnotes fields are then preprocessed to replace those references with appropriate HTML markup.

The styling of the footnotes ckeditor is done via stylesheet added in the `common_design_subtheme`. I couldn't manage to have the stylesheet loaded when added to the `common_design_admin_subtheme` (thanks Drupal).

**Footnotes module**

The preprocessing of the text and footnotes is done in a custom module so that it can also be done when rendering the paragraphs in the forms.

The preprocessing is "simple": extract the references from the text, look for corresponding footnotes, replace the references with html markup and generate the footnotes and append them to the text.

This module contains different templates for the footnote list, items, references and backlinks that can be overridden in the `common_design_subtheme` if needed.

**Footnotes component**

As usual there is a `gho-footnotes` component to style the footnotes and references. There is a bare styling now that will need to be adjusted notably to handle references in Spanish and Arabic.

See the README.md in the `gho-footnotes` component for some TODO.

## Testing

Run `composer update` to install the `ckeditor_config` module then `drush cim` and `drush cr`


html/themes/custom/common_design_subtheme/components/gho-footnotes/README.md
```diff
@@ -0,0 +1,14 @@
...
+Styling for the footnotes and their references in text paragraphs.
+
+See the `gho_footnotes` custom module for the processing and templates.
+
+**TODO**
+
+- The display of the references and backlinks seem to be different in Arabic,
+  notably it looks like the references are not rendered as "super" but normal
+  height in italic (`[lang="ar"] .gho-footnote-reference`).
+- The hit area is really small. This [site](https://wet-boew.github.io/v4.0-ci/demos/footnotes/footnotes-en.html) has nice large references that work well on mobile.
```
rupl, line 13: We can add negative margins to `<sup>` but add padding onto `<a>` thus increasing the hit area without affecting visual aspect. I would additionally absorb a bit of the margin on the initial reading edge (i.e. the left side on LTR, right on RTL). Something like:

```css
sup {
  top: -0.5em;
  left: -.75em;
}
sup a {
  display: inline-block;
  padding: 0 0.5em .25em;
}
```

I think it helps reading rhythm to have the footnote a small bit more snuggly with the preceding text, then we get a more reasonable hit target especially on the number 1 which is quite skinny to begin with. See footnote 1 which has DOM highlighted, plus footnote 2 which is how it would appear normally:

image

I'd rather not add those gray boxes and I doubt Edgar & co. will like the look fo them, having come from a print publication.

html/themes/custom/common_design_subtheme/components/gho-footnotes/README.md
```diff
@@ -0,0 +1,14 @@
...
+- "Enhance" the experience by hiding the footnotes and showing them as tooltips?
```
rupl, line 14: There was this page which I thought was quite a nice UX:

https://www.monde-diplomatique.fr/2020/05/CORDONNIER/61746

rupl (approved):
On first glance this works really well! Having entered my own content to test, I do think it's a reasonable ask on the small team of authors who will be editing the site. Even re-ordering won't be too big of a pain with the simple syntax for creating the footnote reference.

I have added a couple thoughts after reading your TODO list but overall I think this is a great editor foundation and we can enhance presentations in follow-up PRs.

PR #37, opened 2020-10-21
GHO-13: further reading part 2

rupl:
# GHO-13

I closed #34 in favor of this PR. It's now a field with the `link_attributes` thing configured.. but I don't know how to print the information as an independent tag. It renders as a `source` attribute on the `<a>` tag itself... exactly as the module describes. I spent my afternoon trying to extract that link and print it independently, but could not succeed. If you have any ideas let me know. If not, then we have to close this PR and move onto the Paragraphs option.

At a point of desperation, I tried rendering it as a pseudo but the browser is smarter than me and wouldn't let me remove the underline of the pseudo element (makes sense in terms of making deceptive elements that are clickable but look otherwise)

```css
.gho-further-reading a[source]::after {
  content: attr(source);
  text-decoration: none; /* this does nothing, so our design cannot be achieved */
}
```

### DIFF
--- a/html/themes/custom/common_design_subtheme/templates/fields/field--field-further-reading.html.twig
+++ b/html/themes/custom/common_design_subtheme/templates/fields/field--field-further-reading.html.twig
@@ -1,0 +1,86 @@
+{#
+/**
+ * @file
+ * Theme override for a field.
+ *
+ * To override output, copy the "field.html.twig" from the templates directory
+ * to your theme's directory and customize it, just like customizing other
+ * Drupal templates such as page.html.twig or node.html.twig.
+ *
+ * Instead of overriding the theming for all fields, you can also just override
+ * theming for a subset of fields using
+ * @link themeable Theme hook suggestions. @endlink For example,
+ * here are some theme hook suggestions that can be used for a field_foo field
+ * on an article node type:
+ * - field--node--field-foo--article.html.twig
+ * - field--node--field-foo.html.twig
+ * - field--node--article.html.twig
+ * - field--field-foo.html.twig
+ * - field--text-with-summary.html.twig
+ * - field.html.twig
+ *
+ * Available variables:
+ * - attributes: HTML attributes for the containing element.
+ * - label_hidden: Whether to show the field label or not.
+ * - title_attributes: HTML attributes for the title.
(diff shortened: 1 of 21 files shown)


html/themes/custom/common_design_subtheme/templates/fields/field--field-further-reading.html.twig
```diff
@@ -0,0 +1,86 @@
...
+{% else %}
+  <div{{ attributes.addClass(classes) }}>
+    <h3{{ title_attributes.addClass(title_classes) }}>{{ label }}</h3>
+    {% if multiple %}
+      <div class="field__items">
+    {% endif %}
+    {% for item in items %}
+      <div{{ item.attributes.addClass('field__item') }}>
+        {{ item.content }}
+        <span class="source">{# 💰 #}</span>
```
rupl, line 79: here's the prize at the end of the 🌈

orakili (commented):
I created a follow-up PR: https://github.com/UN-OCHA/gho-site/pull/38 to add a custom widget and formatter to handle the source field. The PR is currently against this branch.

rupl:
So with your #38 merged, are we good on this one in your opinion? My local test with your modification was 👍

orakili (changes requested):
Sorry, I still think the chevron should be an icon rather than a character.

VoiceOver on safari reads the character: "comma" (I know...) "the link title":

Screen Shot 2020-10-22 at 18 00 19

and the appearance is font-dependent. Ex: with `courier`:

Screen Shot 2020-10-22 at 18 05 50

So I think we should override the `gho-further-reading-link-formatter.html.twig` in the `common_design_subtheme` and add the following to the markup:

```
<svg class="cd-icon cd-icon--arrow-right" width="16" height="16" aria-hidden="true" focusable="false">
  <use xlink:href="#cd-icon--arrow-right"></use>
</svg>
```

rupl:
Shucks. I was always under the impression that pseudos don't get read by VO or other screen readers, hence why it's not good to put important content in there (e.g. the `source` attribute from previous versions of this PR). But hey if the reading is awkward I can't argue. SVG it is.

orakili (approved):
Works well, thanks.

PR #38, opened 2020-10-22
[GHO-13] Further reading part 2 - custom widget/formatter

orakili:
Ticket: GHO-13

This PR is a complement for https://github.com/UN-OCHA/gho-site/pull/37.

It adds a custom widget and formatter (and template) for the further reading links similar to what was done for the `Dataset` links.

Initially it was going to extend what `link_attributes` does but after looking at the code, it appeared we don't need the module and a simple custom widget would do the trick to add the `source` field. So this PR also removes the `link_attributes` module.

**Form widget**

Screen Shot 2020-10-22 at 13 05 23

Note: there is no placeholder for the source. If we think one should be added, then this can be done by adding a `#placeholder` in the `GhoFurtherReadingLinkWidget.php` file.

**Formatter**

Screen Shot 2020-10-22 at 13 08 38

**Styling/template**

I think that the `›` for the chevron should be replaced by a SVG icon in the markup (`cd-icon--arrow-right`) that could be rotated via CSS on RTL.

There is a default template: `gho-further-reading-link-formatter.html.twig` in the `gho_fields` module that could be overriden in the `common_design_subtheme` for that.

I added some notes in the `components/gho-further-reading/gho-further-reading.css` file about some potential issues with the source.

## Testing

This PR removes the `link_attributes` module so to avoid some issues, I would recommend to first uninstall it before switching to the branch or switch to `develop`, run `drush cim` and then switch to this branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,6 @@
         "drupal/imageapi_optimize_binaries": "^1.0-alpha2",
         "drupal/inline_entity_form": "^1.0-rc8",
         "drupal/layout_paragraphs": "1.0.x-dev",
-        "drupal/link_attributes": "^1.11",
         "drupal/maintenance200": "^1.0",
         "drupal/paragraphs": "^1.12",
         "drupal/paragraphs_viewmode": "^1.0-alpha2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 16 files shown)


rupl:
No need for SVG; the chevron rotates automatically due to ✨  Browser Magic ✨  — I confirmed this by testing all our browsers on the original PR #34, and the styles in #37 are cherry-picked from the old PR.

rupl:
Screen Shot 2020-10-22 at 09 29 25

rupl (approved):
My hero! Thanks for the fixes.

PR #39, opened 2020-10-23
[GHO-17] Make Field story more flexible

orakili:
sTicket: GHO-17

This replaces the `Field Story` content type with a more generic `Story` content type to cater for various types of stories/asides which basically only differ by their "pre-title".

This content type has a new field "Type" which references terms from the new "Story type" vocabulary (ex: Story from the field, Multimedia etc.).

The `image` field has been changed to a more generic `media` field that will accept images, videos and possible embedded content.

**Notes**

"Story type" terms will have to be created via`/admin/structure/taxonomy/manage/story_type/add` so that they can be properly translated. None of the options (autocomplete, inline form) that allow to create terms on the fly provides a way to create a translation of the term inside the node form.

The inline entity form inside the Story paragraph type seems glitchy when translating. I will do a global review of the translatability in a separate PR.

## Testing

This removes a content type, so first step before switching to the branch is to delete existing `Field story` nodes then `drush cim` and `drush cr`.


rupl (approved):
Works great! Nice thinking on the taxonomy for translation. I'd assumed we would use a simple text field but that is a nice way around Paragraphs translation limitations.

PR #40, opened 2020-10-23
GHO-14: Related Article followups

rupl:
# GHO-14

Small follow-ups from work in #33 

- the max-width has been removed from the component, which simplified some responsive image styles config.
- I restored the title which is evident in the PDF where the component is seen in context. We originally built using the screenshot very early in the PDF which had no title.


orakili (approved):
Looks good to me.

PR #41, opened 2020-10-23
GHO-50: Reading width

rupl:
# GHO-50

Adds a max-width to text paragraphs and Further Reading component.


rupl:
The requirements are not as simple as this PR. I think it might be best to cherry-pick any work out of this branch once we have a bit more context, for example in GHO-47

PR #42, opened 2020-10-27
GHO-47: Sub-Articles

rupl:
# GHO-47

A rough cut of the Sub-Article. You can now add Sub-articles to an Article like any other paragraph type. 

The styling of the sub-article isn't finished but it gets the point across for our demo Monday.

I added a new module to allow positioning of the [Title within the Manage Display tab](https://www.drupal.org/project/title_field_for_manage_display). I thought it was the easiest way to position them within _some_ articles but not all, by way of dedicated view modes.

## Testing

```
composer update # new module: title field for manage display
drush cim
drush cr
```


rupl:
@orakili this is ready for review. The styling isn't perfect but if we can get this onto dev for the demo scheduled Monday I'd rather try populating that environment than my local. I get 502s once I add two fully-populated sub-articles and i'm running my drupal container with 4GB which is apparently not enough.

Top comment has been edited to include testing notes like usual

orakili:
@rupl I created a new PR: https://github.com/UN-OCHA/gho-site/pull/46 for the sub-articles, using some of the changes from this one but without the display title module as that could be handled differently via twig templates and a preprocess hook. It also contains a fix for the 502 issue. We probably need to discuss the styling of some elements (hero image notably).

PR #43, opened 2020-10-29
[GHO-10] Needs and requirements

orakili:
Ticket: GHO-10

This adds the needs and requirements figures (top figures).

This creates a new vocabulary with terms matching the overviews and appeals as defined in the column 1 of the "Needs and requirements" spreadsheet. Each term has 3 fields for "people in need", "people targeted" and "requirements (US$)".

A form (/admin/content/needs-and-requirements/import) is provided to import the figures from the spreadsheet (`gho_figures` module).

Adding the figures to an article can be done by adding a "Needs and requirements" paragraph and selecting the appropriate "needs and requirements" term.

This makes the mapping and the styling fairly easy.

This PR also adds a Number formatter in the `gho_fields` module that can handle language aware compact number formatting: `1200000 => 1.2 million`. This follows the rules from the [Unicode Common Locale Data Repository](http://cldr.unicode.org/index/cldr-spec/plural-rules) to ensure the proper translation of the numbers with some GHO specific sauce.

The styling is fairly light:

Screen Shot 2020-10-29 at 18 57 37

Screen Shot 2020-10-29 at 18 23 32

Screen Shot 2020-10-29 at 18 58 32

Screen Shot 2020-10-29 at 18 58 58

## Testing

`composer update`, `drush cim` and `drush cr`.

Sample spreadsheet: 
[GHO2021_NeedsRequirements_2021-test1.xlsx](https://github.com/UN-OCHA/gho-site/files/5457519/GHO2021_NeedsRequirements_2021-test1.xlsx)

### DIFF
--- a/config/field.field.node.article.field_paragraphs.yml
+++ b/config/field.field.node.article.field_paragraphs.yml
@@ -5,13 +5,7 @@ dependencies:
   config:
     - field.storage.node.field_paragraphs
     - node.type.article
-    - paragraphs.paragraphs_type.bottom_figure_row
-    - paragraphs.paragraphs_type.facts_and_figures
-    - paragraphs.paragraphs_type.layout
-    - paragraphs.paragraphs_type.photo_gallery
-    - paragraphs.paragraphs_type.separator
-    - paragraphs.paragraphs_type.story
-    - paragraphs.paragraphs_type.text
+    - paragraphs.paragraphs_type.image_with_text
   module:
     - entity_reference_revisions
 id: node.article.field_paragraphs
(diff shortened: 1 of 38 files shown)


rupl (approved):
Nice, it works very well! The one thing I'd tweak is to somehow prevent word-break in the middle of "millions" — it should either wrap or not in all three cases IMO. This is what I saw when bending the window a bit:

image

It's far from a show-stopper so I'm putting this on dev and we can iterate later on.

config/field.field.node.article.field_paragraphs.yml
```diff
@@ -27,38 +21,35 @@ default_value_callback: ''
 settings:
   handler: 'default:paragraph'
   handler_settings:
-    negate: 0
+    negate: 1
```
rupl, line 24: I would have never thought of this ahead of time but in the future if we flip a boolean like this on a pre-existing list, it's worth calling out in the PR.

I had a serious headache understanding a conflict when merging the new `develop` branch into other in-progress work. Since this `negate` line had not been altered in the other branches it merged without complaint or notification, and that meant I was left trying to understand why all other paragraphs had been disabled since my paragraph types were marked `enabled`, which turned them off. Here's what the conflict looked like:

```
      story:
        weight: 10
<<<<<<< HEAD
      sub_article:
        enabled: true
        weight: 10
=======
        enabled: false
>>>>>>> origin/develop
      text:
        weight: 10
        enabled: false
```

Since I don't want my new feature disabled, I removed the line during conflict resolution. The `enabled: false` actually belongs to `story`, not my new `sub_article` Paragraph type whose branch I was merging `develop` into. Which then broke config in other fun ways 😢

config/field.field.node.article.field_paragraphs.yml
```diff
@@ -5,13 +5,7 @@ dependencies:
...
     - field.storage.node.field_paragraphs
     - node.type.article
-    - paragraphs.paragraphs_type.bottom_figure_row
-    - paragraphs.paragraphs_type.facts_and_figures
-    - paragraphs.paragraphs_type.layout
-    - paragraphs.paragraphs_type.photo_gallery
-    - paragraphs.paragraphs_type.separator
-    - paragraphs.paragraphs_type.story
-    - paragraphs.paragraphs_type.text
+    - paragraphs.paragraphs_type.image_with_text
```
rupl, line 8: also.. is this a bug in drupal? we used `negate` to specify what paragraph types we are NOT using, and so all the stuff marked ENABLED (by way of being `enabled: false` with `negate: 1`) is now unmarked as a dependency... even though those are the dependencies we need.

The only one we **do not** need is `image_with_text` since it is the only Paragraph type **not available** in this field.

config/field.field.node.article.field_paragraphs.yml
rupl: FWIW I do see why you did it. It will be convenient to not have to check each new box we need. I see the negated list as a smart default for our use-case.

orakili: @rupl You're totally right. I did put that in a note when I changed it several days ago but forgot to mention it in the PR. I'm really sorry about that.

PR #44, opened 2020-10-30
GHO-42: webfonts

rupl:
# GHO-42

PR installs the two new webfonts. I originally tried `fontyourface` but that was unsuccessful. I copied the `html.html.twig` into our theme and modified it old-school. We have more important things to worry about.

Any new Sass partials were copied solely to override CD, which specifies fonts in more than one place, requiring us to duplicate the overrides.


orakili (approved):
Looks good.

PR #45, opened 2020-10-30
GHO-10: cosmetic followups

rupl:
# GHO-10

Followup to #43 including these fixes:

- Added the title. Translatable through theme system.
- Adjusted font-sizes at various breakpoints. It may deviate from the design but it does what they will intend for it to do.
- RTL cleanup

I couldn't export myself a new spreadsheet that would upload, so I manually edited the DOM with some more variance in the figures to test all this stuff and make sure it's robust for different amounts of significant digits and the like: 1.4 billion, 42.97 million, etc.


orakili (changes requested):
Looked again at the prototype and in only one place there is something that looks like the title of the "needs and requirements" though I'm more inclined to think it's something else like the title of sub-article:

Screen Shot 2020-11-02 at 15 38 52

On other places, the figures are shown without a title:

Screen Shot 2020-11-02 at 15 39 58
Screen Shot 2020-11-02 at 15 40 04
Screen Shot 2020-11-02 at 15 44 40


So we need to check with Edgar. If they want the possibility to add a title, then I would propose to add a `field_title` the `Needs and requirements` paragraph type and default to a visually hidden "Needs and requirements" when not provided.

Also I'm getting confused with the `em` and `rem`. There seems to be a lack of consistency in our components (my fault probably). For example the `margin-top` and `margin-bottom` of `.gho-needs-and-requirements` were using `em` to reflect the ones from the `.gho-aside`.

rupl:
Ok yeah I guess a title field allows all possible customizations. Good idea.

Regarding the rem and em, to me it is a "standard" convention but there's no official rule. The reason is that using `em` for margin/padding ends up changing those distances if you change the font size. But `rem` is a constant unit as reliable as `px` no matter the context in which it is used.

```html
<h1 class="em">Hello title</h1>
<p class="em">blah blah</p>
<h2 class="rem">Hello sub-title</h2>
<p class="rem">blah blah</p>
```

```css
h1 {
  font-size: 2.5em;
}
h2 {
  font-size: 2em;
}
p {
  font-size: 1em;
}

.em {
  margin: 1em;
}
.rem {
  margin: 1rem;
}
```

The `<h1>` will now have 2.5em of margin on all sides in addition to being quite large text. So if we use margin/padding to line things up, it means additional calculations when tending to font-sizes other than root-level `1em`.

The `<h2>` will have consistent margins with all the paragraphs, which both have a calculated margin of `1rem` because their font-size did not get altered so `1em`/`1rem` are equal in this one context.

image

rupl:
As for branch then yes we can close this, discuss with them, and cherry pick.

PR #47, opened 2020-11-02
[GHO-51] gho-bleed should be loaded by all Article/Sub-Article nodes

rupl:
# GHO-51

Load the `gho-bleed` library very eagerly since it's used by any Article.


orakili (approved):
Thanks!

PR #49, opened 2020-11-02
[GHO-11] Fix bottom figures left/right padding

orakili:
Ticket: GHO-11

According to the PDF, there is no padding left (or right on RTL) for the bottom figures:

PDF:

Screen Shot 2020-11-02 at 19 25 18

PR:

Screen Shot 2020-11-02 at 19 30 55
Screen Shot 2020-11-02 at 19 31 08


rupl (approved):
Worked well on my local and it does indeed look cleaner

PR #50, opened 2020-11-02
GHO-50: reading width pt2

rupl:
# GHO-50

Now that we have sub-articles in place, I re-did the work from #41 to fit within the current content/theme setup.

Testing is just `drush cr` for the libraries


orakili (approved):
Looks good to me.

PR #51, opened 2020-11-02
GHO-50: reading width pt3

rupl:
# GHO-50

A few more cosmetic tweaks to get our components respecting the reading-width column from the design. I will not highlight my shame but let the inline code comments speak for themselves. 🙈


rupl:
Actually I will say one thing: I tested a scenario where a line break was entered into the caption. It survived.

Screen Shot 2020-11-02 at 19 53 17

orakili (changes requested):
Created a follow-up PR to avoid the opening/closing tags in separate templates: https://github.com/UN-OCHA/gho-site/pull/53

Also, according to the latest documents (GHO-CSS.rtf) that Edgar sent, it looks like the max-width should be 720px instead of 820px;

orakili (approved):
I would propose to merge this and tackle all the changes from 820px to 720px in another PR. What do you think?

rupl:
Whoops, too late. I found them all and marked them with identical comments. Shame we can't use anything modern to manage shared values like that, what with IE11 around.

orakili:
> Shame we can't use anything modern to manage shared values like that, what with IE11 around.

We have a css variables ponyfill included in the common_design theme. We could add a variable in https://github.com/UN-OCHA/gho-site/blob/develop/html/themes/custom/common_design_subtheme/sass/cd/_cd-variables.scss for example. What do you think?

rupl:
Yeah ok, let's try it. That can definitely be done in a follow-up

PR #52, opened 2020-11-03
[GHO-55] Use a paragraph type for further reading instead of a field on the node to improved flexibility

orakili:
Ticket: GHO-55

This adds a "further reading" paragraph type to replace the "further reading" field on the node to improve flexibility (like putting further reading links before sub-articles as shown with the Asia and Pacific page in the prototype). This preserves the styling and the form UX is not that different with the field in the paragraph popup:

Screen Shot 2020-11-02 at 19 53 46

### Title

This displays the "futher reading" title as per the prototype PDF v3.

Screen Shot 2020-11-03 at 11 14 44

### Notes

General note: we have to find a way to fix the heading hierarchy for this type of components that can be both in main articles and in sub-articles. When a showing those type of paragraphs in a sub-article we need to change the h2 to h3 for example. Investigating passing some kind of context or using additional view modes. In the worst case, I have something to parse some HTML and fix the heading hierarchy according to the [HTML5 outline algorithm](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines) that should do the trick.

## Testing

`drush cim` and `drush cr`


rupl (approved):
Unsurprisingly this still works as it used to when it was fields. I'm here for the low-drama PR 😁

PR #53, opened 2020-11-03
[GHO-50] Clean and reorganize photo gallery related templates

orakili:
Ticket: GHO-50

Follow-up of https://github.com/UN-OCHA/gho-site/pull/51 to replace 2004 spaghettis with 2020 twig spaghettis ;-)

Also moving around/renaming templates for consistency.

**Note**: this is to be merged into `theming/gho-50-reading-width-pt3`


rupl (approved):
oh this is much nicer and truthfully much more how I'd rather theme a component where most of it is contained and layout can be controlled in one file. Much appreciated for the example!

PR #54, opened 2020-11-03
[GHO-10] Remove workaround for 502 issue as it has nothing to do with the root cause

orakili:
Ticket: GHO-10

This removes the pre-rendering workaround for the 502 issues we has with sub-articles because it's not needed. 

What is causing the 502 is the `x-drupal-cache` header which is really large for complex pages with several sub-articles etc. and cause a "upstream sent too big header" error in the **nginx proxy**. 

Displaying this header is controlled via a setting in the `services.yml` file (ex: shared/settings/services.yml):

```
parameters:
  http.response.debug_cacheability_headers: true
```

Setting that to `false` disables those headers and everything becomes peachy!


rupl (approved):
Following our longer flowdock convo, I still get 502 when logged in as admin and editing a complicated node that has two sub-articles. However, the anonymous page loads and that's good enough for theming so I will let it go until it blocks me again.

PR #58, opened 2020-11-04
Restore admin styles for Facts and Figures paragraph type

rupl:
No ticket, but while auditing the image aspect ratios I noticed these styles got dropped when we set the three-col view mode to  be default.


orakili (approved):
Ah yes, right. Thanks.

PR #59, opened 2020-11-05
[GHO-16] Adjust photo galleries

orakili:
Tickets: GHO-16, GHO-31, GHO-50, GHO-63, GHO-67

**Note:** I think we need to have a discussion about the image styles and responsive image styles asap.

This PR is to adjust the photo galleries to make them more consistent (gho-aside) with the components added after that and take into account other recent changes. So:

### 1. GHO-31, GHO-63: new attempt at a convention for the image styles

This introduces new image styles that I hope are easy to reason with and maintain.

They are all proportional to the full max width of the content region (1140px) with an indication of their ratio and are generic enough to be easily re-used in different responsive image styles.

Examples: `Full width - 16:9 - 100%`, `Full width - 16:9 - 50%`, `Full width - 16:9 - 33%`, `Full width - 16:9 - 200%`

Screen Shot 2020-11-05 at 12 44 07

### 2. GHO-16: new view modes/responsive image styles

I think it's better to differentiate by component at the responsive image style level than the image style level.

So this introduces 2 new responsive image styles for photo galleries and associated media view modes: `Photo gallery - Single column` and `Photo gallery - Two columns`.

I also removed the manual selection of the view mode for the photo gallery. It's now automatically selected based on the number of photos.

### 3. GHO-67: photo credits removed

This removes the photo credits on each individual photo.

### 4. GHO-50: Reading width

This introduces a `--reading-width` css variable (set to 720px) in `sass/base/base.scss`.

**Note:** I applied this contraint to the entire photo gallery in two-columns mode for Edgar review as the spanning to the entire width we had was different than what is in the PDF anyway and in my opinion didn't look good.

**Note:** the `photo gallery` responsive image styles should be adjusted based on that but as this width, margins etc. keep changing, they are now using a safe default. We can have an optimization pass after everything else is done.

### Testing

`drush cim` and `drush cr` should be enough.


rupl (approved):
I pointed out the issue with IE11, but I've found the cause and will prepare a follow-up for that. I also have a fix for the caption spacing in the branch.

Otherwise I do like the convention and for now I'm happy to stick with it. Indeed you are right that the component namespacing probably makes more sense inside the responsive image style, and not the Drupal image styles themselves. The generic preparation of grid-like image styles combined with "semantic" groupings inside the responsive image style config is a nice blend of reusability and concreteness.

I don't like how IE11 has layout shifts due to the CSS custom props polyfill, but no Gallery seems to be above the fold so for now we can go with it. However if we see this happening on the text paragraphs, it might be worth adding an IE11 fallback back in there.

PR #60, opened 2020-11-05
[GHO-16] photo gallery pt3

rupl:
# GHO-16

Follow-ups for #59 

- IE11 photo grallery layout fixed. [Details if you want them](https://stackoverflow.com/a/52669657).
- Location/Caption had space re-introduced, I presume when we shuffled the markup around and stopped using `display:inline` on caption directly.


orakili (approved):
Thanks.

PR #61, opened 2020-11-05
[GHO-33] Adjust story content and type and allow video embeds

orakili:
Ticket: GHO-33, GHO-63

This PR adjusts the story, notably extracting the credits from the media and displaying it at the end of the text and add a video media type that can be used by the story.

**Note:** The text field for a story is "formatted" (bold, itatlic, links) as opposed to the caption of a photo gallery which is just plain text, to accommodate for more "complicated" texts. As a result the content of the story `field_text` is a list of `<p>` so we wrap it and the credits inside a `div` and not a `p` like for the photo galleries.

Screen Shot 2020-11-05 at 21 59 05

**Note:** The videos also have a mandatory credits field.

**Note:** There is a new "Story" responsive image style which uses the Full width 100% image style while it's currently limited to 720px. This is on purpose and to be reviewed during the optimization phase in case the reading width gets changed in the mean time.

## Testing

`drush cim` and `drush cr`


rupl (commented):
What, no footnotes in the captions?? 😁 

Checks out and the new image style is exactly the minimum that we sent to Edgar this morning. Nice.

The one thing I did notice is that when creating a new Video within the context of an Article, I was not prompted to add the Credit field, nor did it prevent me from saving the Video entity and selecting it for the Story entity. Screenshot of initial Video creation in context (Article node > Story paragraph > Video entity)

image

When I directly edited the Video entity (or created a new Video as an independent task) then I do get the required Credit field, and cannot save the Video without supplying a value.

Screen Shot 2020-11-05 at 17 39 14

rupl (approved):
I double-checked every field setting in the entire chain and don't see anything out of place. Most notably, the Credits field is marked as required, and its Manage Form Display settings reflect exactly what you'd expect in the second screenshot (when independently adding/editing a Video entity outside the context of Paragraphs).

The best clue I could dig up is in the `media_library` module (found in  `html/core/modules/media_library/src/Form/OEmbedForm.php` in our repo) and the only field it presents to create the OEmbed is the video url.

I'm tentatively marking this as approved and if you're not aware of a hook that can be used to surface fields other than "Name" (which appears after entering the YouTube URL and then clicking Add) we might just have to document that Video Credits have to be added after the fact. When the field is finally filled out, it displays exactly as intended:

image

rupl:
booyah! of course I had to submit my comment before finding it, but the Media Library form view-mode of the Video entity needed the field hoisted. PR against this branch: #63

PR #62, opened 2020-11-05
GHO-46: translations

rupl:
# GHO-46

I had to copy some templates into our CD subtheme to get Drupal to recognize the translation strings. The PO files are in a new top-level directory called `translations` outside the web root. I thought that it's good to keep a copy around in case we need to re-import.

I also imagined the possibility that maybe we can auto-import the PO files when deploying a new copy of the site. I didn't look into how exactly that would be done, or if it's even a good idea.


rupl:
I've imported the PO files from this branch into dev. However the branch has to be merged and deployed before the templates in the PR can display as translated on dev.

image

orakili (approved):
Looks good to me. Thanks.

PR #63, opened 2020-11-05
GHO-33: multimedia credit field

rupl:
# GHO-33

Small improvements to the work in #61 to surface the credit field, and also mark the Image/Video as required when creating a Story entity. The config screen I finally found:

image

And the new upload form with required Credits field:

image


orakili (approved):
Thanks for fix. As for the media field being mandatory, it was initially not required to allow having stories without a visual. We can see if that happens and adjust accordingly.

PR #64, opened 2020-11-06
[GHO-57] Highlight node status for editors

orakili:
Ticket: GHO-57

This PR adds an attribute to the nodes for editors to display the language and publication status. This, in my opinion, helps a lot to make sense to what is displayed on a page:

Screen Shot 2020-11-06 at 12 32 27


rupl (approved):
Nice, works as advertised.

PR #65, opened 2020-11-06
[GHO-57] Use simple autocomplete to select a story

orakili:
Ticket: GHO-57

Use simple autocomplete to select a story instead of an inline entity form.


rupl (approved):
easy-peasy

PR #66, opened 2020-11-06
[GHO-73] Add footnotes and sources fields to stories

orakili:
Ticket: GHO-73

This adds the footnotes and sources fields to the stories to handle case like the "Education in emergencies: the lingering effects of COVID-19" story in the Section 1 which contains both footnotes and sources.

For the footnotes, this re-uses the same process as the text paragraph.

The sources are a bit similar in idea, using a textarea with ckeditor to allow limited formatting.

Screen Shot 2020-11-06 at 15 01 03

## Testing

`drush cim`, `drush cr` and create or edit a story and footnotes and sources.


rupl (approved):
Works great :shipit:

PR #67, opened 2020-11-06
[GHO-66] Facts and figures credits

orakili:
Tickets: GHO-66, GHO-63

This displays the credits of the images of the facts and figures after the texts like done for the stories and photo galleries.

It also uses a new "Facts and figures" responsive image style and associated media and paragraph view modes.

Screen Shot 2020-11-06 at 16 27 43

I also removed all the media and paragraph view modes that we are not using anymore for clarity. I'll do the same for the responsive image styles and image styles in another PR.

## Testing

`drush cim` and `drush cr` should be enough


config/core.entity_form_display.taxonomy_term.section.default.yml
```diff
@@ -34,6 +34,11 @@ content:
     weight: 2
     region: content
     third_party_settings: {  }
+  translation:
+    weight: 10
+    settings: {  }
+    third_party_settings: {  }
+    region: content
```
rupl, line 41: yay to this. I've somehow been battling this piece of config for a day or two and couldn't find the source.

html/themes/custom/common_design_subtheme/templates/paragraphs/paragraph--image-with-text--facts-and-figures.html.twig
```diff
@@ -0,0 +1,62 @@
+{#
+/**
+ * @file
+ * Theme implementation for an image with text in Facts and figures view mode.
+ *
+ * Overrides paragraphs/templates/paragraph.html.twig.
```
rupl, line 6: You are adding these manually now, right? It's nice to know where it came from. For the record, whenever I copy `field.html.twig` it has always been from `classy` theme.

html/themes/custom/common_design_subtheme/templates/paragraphs/paragraph--image-with-text--facts-and-figures.html.twig
```diff
@@ -0,0 +1,62 @@
...
+%}
+{% block paragraph %}
+  <div{{ attributes.addClass(classes) }}>
+    {% block content %}
+    {{ content.field_image }}
+    <div class="gho-facts-and-figures__caption caption">
+      {{ content.field_text }}
+      <span class="gho-facts-and-figures__credits credits">{{ content.credits }}</span>
+    </div>
+    {{ content|without(['field_image', 'field_text', 'credits']) }}
```
rupl, line 59: Is this just to safeguard a future scenario where we output some other piece of content? It should produce nothing at the moment, correct?

rupl (approved):
Looking slick. I left some questions inline for my own curiosity but it looks great when testing.

html/themes/custom/common_design_subtheme/templates/paragraphs/paragraph--image-with-text--facts-and-figures.html.twig
orakili: Yep, correct, I copied that from the changes to the stories where I added the footnotes and sources. So in case we add something to the facts and figures then it should appear there and prevent some headache.

html/themes/custom/common_design_subtheme/templates/paragraphs/paragraph--image-with-text--facts-and-figures.html.twig
orakili: Yeah, it's not consistent. I also generally copy from classy as well but in the case of the paragraph templates they are copied from the paragraphs module. In an ideal world we would have a proper header in all our templates but well...

rupl: Well I like the idea. I will try to follow the convention from here on out.

PR #68, opened 2020-11-06
[GHO-75] Remove related articles from articles as that will generated automatically

orakili:
Ticket: GHO-75

This tiny PR is to prepare for the automation of the "more from this section" component, that will also be used to populate the homepage. So for that, we first remove the "related articles" field form the articles. We'll generate the list from the menus.

## Testing

`drush cim` and `drush cr`


rupl (approved):
Another easy one. :shipit:

rupl:
On second thought, I am not immediately merging this because I imagine people might want to be able to see the Related Articles on dev in order to present to their stakeholders. And I imagine they'll want to be able to see that regardless of how it works in the background, so perhaps we can hold onto this until the other mechanism is in place.

rupl:
Ok with the other work merged, I had to manually resolve some merge conflicts but in all five cases, it seemed to be a matter of retaining the new PDF download field `field_report_pdf` while deleting config for the field this branch removes: `field_related_articles`

rupl:
Removing approval bc a local test indicated I either merged incorrectly or there's still a bit of config that needs to be removed:

```
 [error]  Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization.
Configuration <em class="placeholder">core.entity_view_display.node.article.home_page</em> depends on the <em class="placeholder">field.field.node.article.field_related_articles</em> configuration that will not exist after import. in Drupal\Core\Config\ConfigImporter->validate() (line 755 of /srv/www/html/core/lib/Drupal/Core/Config/ConfigImporter.php).

In ConfigImportCommands.php line 357:

  The import failed due to the following reasons:
  Configuration <em class="placeholder">core.entity_view_display.node.article.home_page</em> depends on
  the <em class="placeholder">field.field.node.article.field_related_articles</em> configuration that
  will not exist after import.
```

rupl:
Ok looking good after one more manual removal. My bad on the merge.

PR #69, opened 2020-11-06
[GHO-30] Add a report link field to articles and a new Home page view mode

orakili:
Ticket: GHO-30

Preparation for the report download link: add a link field to the articles and a "home page" view mode for the home page. That will be set automatically via a hook for the home page (node/1) in another PR.

## Testing

`drush cim` and `drush cr`


config/core.entity_form_display.taxonomy_term.section.default.yml
```diff
@@ -34,6 +34,11 @@ content:
     weight: 2
     region: content
     third_party_settings: {  }
+  translation:
+    weight: 10
+    settings: {  }
+    third_party_settings: {  }
+    region: content
```
rupl, line 41: whyyy does this thing keep showing up??

rupl (approved):
Link is appearing when I add it to the homepage. No styling to speak of but I imagine we're waiting til we get a final design?

config/core.entity_form_display.taxonomy_term.section.default.yml
orakili: I don't know. It seems to show up when touching the view modes.

PR #70, opened 2020-11-06
GHO-46: more translations

rupl:
# GHO-46

Another round of translations. Javi also caught a couple errors in our source strings. I've filed CD-235 to fix the `UN-OCHA` upstream but in the mean time updated the `common_design_subtheme` to fix the problem on GHO.


orakili (approved):
Thanks.

PR #71, opened 2020-11-06
[GHO-75] Section index

orakili:
Ticket: GHO-75

This PR introduces a Section index paragraph type to be used to build the homepage as well as some field formatter/widget and a preprocessor to automatically generate lists of related articles using the drupal menus.

### Section index paragraph

This paragraph is normally to be used only to build the homepage and contains a title, an optional pre-title, description and menu entry (one of the direct children of the "main" navigation menu (ex: introduction etc.)).

Screen Shot 2020-11-07 at 0 02 50

### Related articles - Logic

The same logic (in `GhoRelatedArticlesFormatter`) is used for both the list of articles in the section index and the "more from this section" at the end of each article. 

For the section index, it gets the menu links for the chosen main menu navigation item (direct child of the "main" menu, ex: Introduction) and retrieve the children of that menu entry and the associated nodes and display them in the "related article" view mode.

For the "more from this section", there is preprocessor for article nodes that retrieve the associated menu link if any, and so the same thing as above with the parent of this link which is one of the direct children of the "main" menu. In that case, the article node being viewed is excluded from the list of related articles.

### Styling

The same `gho-related-articles` component is used for the list of articles in the section index or the "more from this section".

There is also a `gho-section-index` component to style the paragraph pre-title, title and description.

I refactored a bit the related articles, to have the image above the title on small screens, have the "read article" aligned at the bottom and use a svg icon for the arrow. Tested with browserstack and it was working well in IE and other browsers:

**Section index**

Screen Shot 2020-11-07 at 22 32 27

**More from this section**

Screen Shot 2020-11-07 at 21 53 37

**Small screen**

Screen Shot 2020-11-07 at 21 53 48

**RTL**

Screen Shot 2020-11-07 at 22 01 02

### Image style

The previous responsive image style for the related articles was replaced with a new one following the recent convention and using the 16:9 image styles.

## Testing

`drush cim`, `drush cr` and add a Section index paragraph.


orakili:
For the "More from this section", actually there is no need for an additional field because we can get the menu link info from the node and use some code similar to the `GhoMenuFormatter.php` to add the list of related articles via a `hook_preprocess_node__article()` in the `common_design_theme.theme` for example. (And this way exclude the current article from the list.)

orakili:
We had a terrible weather today and had to cancel our plans so I used a few hours to finish this work. I've updated the description of the PR.

rupl:
I'm getting what seems to be a one-time error when importing config:

```
[error]  Drupal\Core\Config\ConfigException: Errors occurred during import in Drush\Drupal\Commands\config\ConfigImportCommands->doImport() (line 343 of /srv/www/vendor/drush/drush/src/Drupal/Commands/config/ConfigImportCommands.php).

In ConfigImportCommands.php line 357:

  The import failed due to the following reasons:
  Unexpected error during import with operation create for core.entity_view_display.paragraph.section_
  index.default: The &quot;gho_related_articles&quot; plugin does not exist. Valid plugin IDs for Drup
  al\Core\Field\FormatterPluginManager are: [long list not containing gho_related_articles]
```

When I `drush cr` and try again, it doesn't produce the error anymore.

rupl (approved):
I could never reproduce the error I mentioned, and all the other frontend bits check out for me: top separator, image sizing is reasonable enough, separator between articles.

I noticed that some Sub-Articles don't seem to have images associated and we probably need a follow-up to handle situations where no thumbnail exists. It might also just be that the huge library of images we have available will be employed to provide thumbnails, since it's not portrayed in the PDF how to handle a missing image.

PR #73, opened 2020-11-09
[OPS-6971] Allow HID logins during maintenance mode, so we can test and populate prod with content.

cafuego:
This should maybe eventually go into social_auth_hid with a setting to turn it off or on.


orakili (approved):
Working well.

PR #79, opened 2020-11-10
[GHO-25] Language visibility

orakili:
Ticket: GHO-25

**Background**: 

1. Drupal's default behavior is to display content in their default language when the translation is missing. So we end up, for example, with articles in Arabic with sub-articles in English because not yet translated.
2. It's been requested to be able to only show the homepage in Arabic for example while content in that language is being added to the site.

**Solution**

So this modules deals with dark magic to add access checks to allow the visibility of content in a given language (after a million of hours battling Drupal):

1. It defines a "view untranslated content" permission (given to admins) to allow to see content in any language regardless of the status of the translations. Said otherwise, if you have this permission, then you get the default Drupal's behavior.
2. It repurposes the "promote" flag for the homepage to control the visibility of  content in the language of the homepage across the site. Said otherwise content in Arabic is only displayed if the Arabic translation of the homepage is `published` and `promoted`.
3. It provides a route access handler to show a `404` when viewing a page in a language whose visibility is not enabled (access denied but 404 instead of 403 for better UX)
4. It provides some logic to deny access to entities if they don't have a translation in the current language or the global visibility of the language is not enabled (promote of the homepage).
5. Additionally and could have been in another PR, it allows admins to translate any entity.

### Testing

`Drush cim`, `drush cr` then as an anonymous visitor try to view pages in a language for which there is no translation or for which the homepage in the same language is not published and promoted. Then add a translation if none and promote/publish the homepage in that language. Do the same when logged in as normal admin user (not user 1) and check you have access to everything.

Also, as an anonymous user, on a Arabic page (and Arabic homepage published/promoted) with a sub-article not translated, check that the sub-article doesn't appear.

### DIFF
--- a/config/core.entity_form_display.node.article.home_page.yml
+++ b/config/core.entity_form_display.node.article.home_page.yml
@@ -1,0 +1,96 @@
+uuid: 713e381c-11b6-4056-ab08-9ccb00957e26
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_form_mode.node.home_page
+    - field.field.node.article.field_appeals
+    - field.field.node.article.field_author
+    - field.field.node.article.field_hero_image
+    - field.field.node.article.field_paragraphs
+    - field.field.node.article.field_report_link
+    - field.field.node.article.field_section
+    - field.field.node.article.field_summary
+    - field.field.node.article.field_thumbnail_image
+    - node.type.article
+  module:
+    - layout_paragraphs
+    - link
+    - media_library
+    - text
+id: node.article.home_page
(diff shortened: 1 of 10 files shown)


rupl (approved):
Once I got over the initial panic of _everything breaking_ and checked the "display this language" box for english, it all went really well 😆 

I won't even pretend to review the access code. Maybe if you really want a second opinion get @attiks in here to take a glance?

Some bidi issues with navigation when Arabic menu not yet translated, but by editing DOM confirmed it will all fall back into place once translated.

PR #80, opened 2020-11-10
[GHO-19] Interactive content

orakili:
Tickets: GHO-19, GHO-21

This adds a paragraph type for interactive content with a type (pre-title), title, image (screenshot of the interactive content), link (to the interactive content) and dataset source.

This quite similar to the "Stories" except it's a paragraph type not a content type because we don't need to have individually accessible pages for those as they embed/link to external content.

There is new "interactive content type" vocabulary to avoid mixing the terms with the story types.

Screen Shot 2020-11-10 at 20 22 11

This uses the `linked_responsive_image_media_formatter` custom module to generate the clickable image.

The  styling re-uses the story  type one with a large image (904px).

Screen Shot 2020-11-10 at 19 41 20

### Testing

`drush cim`, `drush cr` and add some interactive content type terms and then a interactive content paragraph to an article.

### DIFF
--- a/config/core.entity_form_display.paragraph.interactive_content.default.yml
+++ b/config/core.entity_form_display.paragraph.interactive_content.default.yml
@@ -1,0 +1,60 @@
+uuid: 39428b73-cb1d-4f07-bb53-201cf088327c
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.paragraph.interactive_content.field_dataset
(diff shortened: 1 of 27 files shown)


rupl (approved):
alt of that image gonna get _crowded_ but I suppose as a paragraph type it won't be a big pivot to add another text field for a proper "caption" of the graphic.

I see a 720px image (lined up with reading-width text) but also a small thing to change if needed.

PR #82, opened 2020-11-10
GHO-30: style home page

rupl:
# GHO-30

Theming for home page, including PDF link. I used the Summary to print out the custom message. I felt it was safe since Home Page will never be included in the Section Overviews (which use Summary to output the block of text) and to avoid another special-use field that will just cause confusion on all other Articles. Using paragraphs also seemed like it made things far too complicated, as they'll be used for normal-style content in the future (if the design PDF is to be believed)

The new image styles and responsive image style were discussed in flowdock and I did end up adding a namespace of sorts to keep them well-grouped.

I briefly messed around with a `font-size: calc(em + vw)` which results in a far-smoother progression. Pair it with a `max-width: ch` value (which also gradually increases with `font-size` and you have a far superior solution that looks great on all screens... if we can get sign-off away from specific px-values.

Speaking of the pixel values, I actually couldn't track them down when looking through ticket backlog so I went with my gut and we can fix them once I find the specs again.

Also included were some core language updates.


rupl:
After merging #80 with the new image styles I wanted to highlight the reason I'd namespaced mine with the word "uncropped" best illustrated by screenshotting the two branches merged together resulting in the following dropdown:

Screen Shot 2020-11-10 at 17 58 22

The alphabetization of various numbers which correspond to different tokens within our image style titles (100, 16, 180) ends up separating one logical group of styles into two groups within the dropdown. Even IF we fixed this, it STILL wouldn't alphabetize to my satisfaction since 100% comes before 50% and so forth.

The list is sorted by the machine name of the style, not the title. Just a thing to keep in mind when we extract all this from GHO and into CD or another project, distro, or starterkit.

rupl:
I dropped all my new image styles and adopted the ones from #80 with a couple additions that I needed, but using the convention of the ones that existed.

orakili (approved):
Looks good but there are some display issues on mobile with the margins notably:

Screen Shot 2020-11-11 at 15 08 20


I'll deal with those when tackling GHO-79.

rupl:
Oh man, you're right I just slapped those margins on the top and bottom and never noticed when testing mobile... I was so focused on the title

PR #83, opened 2020-11-10
GHO-60: update webfonts / titles

rupl:
# GHO-60

Drop DM Serif, add Roboto 900. Titles were already 52px across the board. Updating the homepage PR #82 to adopt the correct font-size since this was the ticket I mentioned not being able to find in that PR description.


orakili (approved):
Working as advertised, thanks.

PR #87, opened 2020-11-12
[GHO-81] Accumulated footnotes

orakili:
Ticket: GHO-81

This implements a new mode for the footnotes that accumulates them at the bottom of a node page and also with the fancy popup a la https://www.monde-diplomatique.fr/2020/05/CORDONNIER/61746

I added a new "preview" view mode for some nodes and paragraphs that is used in the forms so that we can still display the footnotes below the text for easier review. This should not change anything to what was displayed before in the forms.

**Note:** The behavior is to show a footnote in the popup if it is in the top half of the viewport. This goes with the max-height of 50% for the popup so that it doesn't display a footnote that is hidden by the popup. We could adjust a bit like 60/40 but it my tests with a paragraph with lots of references that was the best experience.

I managed to find  a way to have the footnotes popup fixed at the bottom of the page while spanning to the left and right border as requested in prototype v4, that seems in all browsers including IE11:

**prototype**

Screen Shot 2020-11-12 at 17 54 54

**Implementation**

Screen Shot 2020-11-12 at 17 55 25

**Note:** that is goes above the grey margins on smaller screens but I think that's ok and that could possibly be changed by subtracting the margin width to the translate, width and padding used for this effect in gho-footnotes.css:

Screen Shot 2020-11-12 at 17 58 08

**Script**

The script for the popup is plain vanilla javascript using intersection observers when available or a scroll event otherwise. I tried to optimize a bit the scroll event and I didn't experience any noticeable slow down when the script is enabled.

## Note

The styling of the numbers of the footnotes differs from the prototype v4 in which they seem to have the number be a backlink to the reference. I'll change that in another PR after confirming that they will never have multiple references to a single footnotes. I also find that having the backlink at the beginning of the footnote is not a great experience when using a screenreader or navigating with a keyboard, so maybe we don't need to change it.

## Testing

`drush cim`, `drush cr` and visit a page with footnotes.

### DIFF
--- a/config/core.entity_form_display.node.article.default.yml
+++ b/config/core.entity_form_display.node.article.default.yml
@@ -5,6 +5,7 @@ dependencies:
   config:
     - field.field.node.article.field_appeals
     - field.field.node.article.field_author
+    - field.field.node.article.field_caption
     - field.field.node.article.field_hero_image
     - field.field.node.article.field_paragraphs
     - field.field.node.article.field_report_link
@@ -13,6 +14,7 @@ dependencies:
     - field.field.node.article.field_thumbnail_image
     - node.type.article
   module:
+    - double_field
     - inline_entity_form
     - layout_paragraphs
     - link
@@ -47,6 +49,33 @@ content:
       allow_duplicate: false
     third_party_settings: {  }
     region: content
+  field_caption:
+    weight: 6
+    settings:
+      first:
+        type: textfield
+        label_display: block
+        size: 60
+        placeholder: 'Ex: Mbuji-Mayi, RDC'
+        label: Ok
+        cols: 10
+        rows: 5
+        prefix: ''
+        suffix: ''
+      second:
+        type: textarea
+        label_display: block
+        size: 10
+        placeholder: ''
+        label: Ok
+        cols: 10
+        rows: 5
+        prefix: ''
+        suffix: ''
+      inline: false
+    third_party_settings: {  }
+    type: double_field
+    region: content
   field_hero_image:
     type: media_library_widget
     weight: 5
@@ -55,7 +84,7 @@ content:
     third_party_settings: {  }
     region: content
   field_paragraphs:
-    weight: 8
+    weight: 9
     settings:
       preview_view_mode: preview
       nesting_depth: 4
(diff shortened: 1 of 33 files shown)


rupl:
When I set everything up I'm seeing some strange effects and no pop-up at all. I got a clean config import and never saw any errors when running server commands.

When I view as both admin and visitor, I no longer see the footnotes at all.

- I checked Permissions (even though it wouldn't explain Admin missing the footnotes) and there is an unchecked perm for the Administrator role to use the Footnotes text input
- I tried viewing both an Article containing a Sub-Article with footnotes, and viewing that Sub-Article directly as a top-level page. Same behavior in both cases.

In desperation, I tried searching both HTML source and the browser DOM for the elements in question and they seem to be missing. I tried loading with JS disabled too. No dice. The only evidence I found that the footnotes should appear is the following twig debug when directly viewing HTML source:

```html
<!-- THEME DEBUG -->
<!-- THEME HOOK: 'gho_footnote_list__accumulated' -->
<!-- FILE NAME SUGGESTIONS:
x gho-footnote-list--accumulated.html.twig
x gho-footnote-list--accumulated.html.twig
* gho-footnote-list.html.twig
-->
<!-- BEGIN OUTPUT from 'themes/custom/common_design_subtheme/templates/content/gho-footnote-list--accumulated.html.twig' -->
 
<!-- END OUTPUT from 'themes/custom/common_design_subtheme/templates/content/gho-footnote-list--accumulated.html.twig' -->
```

My best guess is that the `footnotes` var inside `gho-footnote-list--accumulated.html.twig` is empty. it appeared to be when I dumped form the template, producing the string `NULL` within the begin/end comments I pasted above. Since the library is attached only when `footnotes` exists, that's why none of the JS is loading and executing.

In the process of doing all this I very carefully checked the names of the preprocess hooks and corresponding template names and so forth.. I couldn't find a problem and honestly did not expect to, given that you submitted the PR and I imagine it was working for you.

I tried resaving the Text Paragraphs which contained the footnotes, but it doesn't seem to affect my local's behavior. Perhaps we can look together tomorrow?

rupl:
I see the permission issue is addressed in #88

rupl (approved):
Ok chalk it up to docker troubles? Once we reviewed together on the call it worked absolutely perfect and beautiful and stuff.

@orakili I'm approving and see that the merge conflict looks pretty simple, but will leave it to you to clean up since it's two of your branches.

PR #89, opened 2020-11-12
[GHO-34] Article caption

orakili:
Ticket: GHO-9, GHO-34

This adds the article caption via a double field on the node and using the credits from the hero image.

Screen Shot 2020-11-12 at 22 44 46

Screen Shot 2020-11-12 at 22 36 04

**Note** The margin between the title and the caption need to be adjusted as well as lots of other places so leaving that for another PR.

## Testing

`drush cim`, `drush cr` and add a location/caption to an article.

Note: there may be a classic issue of the chicken and egg like it happened in other PRs, when running the drush commands. Another run of those 2 commands should fix it.

### DIFF
--- a/config/core.entity_form_display.node.article.default.yml
+++ b/config/core.entity_form_display.node.article.default.yml
@@ -5,6 +5,7 @@ dependencies:
   config:
     - field.field.node.article.field_appeals
     - field.field.node.article.field_author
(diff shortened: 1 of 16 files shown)


rupl (approved):
I did see the plugin error but running the config import after cache clear made it all fall into place. Nice job with the image credit.

PR #90, opened 2020-11-12
[GHO-99] Menus

orakili:
Ticket: GHO-99, GHO-85

Allow editors to modify menus  via structure > menus and enable menu translations.

`drush cim`, `drush cr`, log in as normal admin user, go to "structure > menus", click on edit menu for "main menu", check you can re-order, rename and translate sub-menus.

### DIFF
--- a/config/core.base_field_override.menu_link_content.menu_link_content.changed.yml
+++ b/config/core.base_field_override.menu_link_content.menu_link_content.changed.yml
@@ -1,0 +1,18 @@
+uuid: 3864ae5f-9f4b-4af1-9b35-96c31f1b0538
+langcode: en
+status: true
+dependencies:
+  module:
+    - menu_link_content
+id: menu_link_content.menu_link_content.changed
+field_name: changed
+entity_type: menu_link_content
+bundle: menu_link_content
+label: Changed
+description: 'The time that the menu link was last edited.'
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: changed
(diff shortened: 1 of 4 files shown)


rupl (commented):
Reordering menu items was no problem, but translating a menu item I got this error:

```
Error:
Non-translatable fields can only be changed when updating the original language.
```

The form is quite short and even before saving displays a warning:

```
Warning:
Fields that apply to all languages are hidden to avoid conflicting changes.
```

I looked through permissions and did see one that might be related: "Translate custom menu link" but enabling did not change behavior, and actually testing with user/1 produces the same error.

rupl:
The other field on the Translate form is **Parent link** which I imagine wouldn't need to be translated. Is there some way of choosing what gets translated and what gets omitted from this form?

rupl:
The menu item translation is successful as user/1 — reviewing perms one more time to see if something sticks out

rupl:
Ok a more detailed report of what I've seen as I posted all the short comments above:

When I click **Edit Menu** and then go to Translate an individual menu link, **I am successful for top-level links only** for both user/1 and the `administrator` role. I was able to translate _Introduction_ into all four languages: two translations made by a user with `administrator` role, and one by user/1.

When I try to Translate a second-level menu link, both user/1 and `administrator` display the error I initially noted above:

> Non-translatable fields can only be changed when updating the original language.

orakili:
@rupl The issue was with a setting in https://gho.test/admin/config/regional/content-language to hide the non-translatable fields on the form. Apparently Drupal doesn't like itself...

rupl:
Good to know about the settings! Yeah I was looking for more of a "Manage form display" type tab but obviously it's not there.

rupl:
Yes now there are many more form items on the translation page and second-level menu items are translating as expected.

rupl (approved):

PR #91, opened 2020-11-13
[GHO-101] Allow 2 columns layout and limit width

orakili:
Ticket: GHO-101

This brings some simple changes to allow the use of the 2 columns layout. That doesn't change anything for the existing content but in case it is decided to go with the 2 columns option then it's "ready".

I also removed the other layouts (3, 4 columns, grid) to avoid giving "ideas".

`drush cim`, `drush cr`

### DIFF
--- a/config/core.entity_view_display.node.article.default.yml
+++ b/config/core.entity_view_display.node.article.default.yml
@@ -13,7 +13,7 @@ dependencies:
     - field.field.node.article.field_thumbnail_image
     - node.type.article
   module:
-    - entity_reference_revisions
+    - layout_paragraphs
     - link
     - user
 id: node.article.default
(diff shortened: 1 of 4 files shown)


config/core.entity_view_display.node.article.default.yml
```diff
@@ -14,6 +14,8 @@ dependencies:
     - field.field.node.article.field_thumbnail_image
     - node.type.article
   module:
+    - entity_reference_revisions
```
rupl, line 17: @orakili when I read the rest of the diff it seems we can rid ourselves of the `entity_reference_revisions` dependency. I included it in this conflict resolution. I don't think it's hurting anything but can you take a look and confirm?

orakili: It's removed in `config/core.entity_view_display.node.article.home_page.yml` and it's a dependency for the `entity_reference_revisions_entity_view` formatter that is replaced by the `layout_paragraphs` one so, yes, I think we can remove it.

PR #93, opened 2020-11-13
[GHO-97] Remove black background for interactive content

orakili:
Ticket: GHO-97

As simple as what the title says.


rupl (approved):
:shipit:

PR #94, opened 2020-11-13
[GHO-98] Add highlight style for paragraphs in ckeditor

orakili:
Ticket: GHO-98

Add a "highlight" for paragraph (`<p>`) in ckeditor for `text` paragraph types to allow the following:

Screen Shot 2020-11-13 at 13 08 37

`drush cim`, `drush cr`, edit a `text` paragraph, select a paragraph, click "styles" dropdown and select "Highlight".

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/gho-text/gho-text.css
+++ b/html/themes/custom/common_design_subtheme/components/gho-text/gho-text.css
@@ -1,3 +1,8 @@
 .paragraph--type--text {
-  max-width: 720px; /* reading-width */
+  max-width: var(--reading-width);
+}
+.paragraph--type--text p.highlight {
+  font-size: 1.5rem;
+  font-weight: 700;
+  line-height: 2.25rem;
 }
--- a/html/themes/custom/common_design_subtheme/components/gho-text/gho-text.css
+++ b/html/themes/custom/common_design_subtheme/components/gho-text/gho-text.css
@@ -1,7 +1,7 @@
 .paragraph--type--text {
   max-width: var(--reading-width);
 }
-.paragraph--type--text p.highlight {
+.field--name-field-text .highlight {
   font-size: 1.5rem;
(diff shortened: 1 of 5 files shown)


html/themes/custom/common_design_subtheme/components/gho-text/gho-text.css
```diff
@@ -1,3 +1,8 @@
 .paragraph--type--text {
-  max-width: 720px; /* reading-width */
+  max-width: var(--reading-width);
+}
+.paragraph--type--text p.highlight {
```
rupl, line 4: I randomly tried to add the Highlight style to the top Paragraph of the Section Overview, and CKEditor shows it but the public website doesn't because its selector was:

```

```

I think we can change to this and catch all of them:

```suggestion
.field--name-field-text .highlight {
```

Using ☝️ also works to make it slightly more wysiwyg in the editing screen as well.

rupl: Admin styles added to #98  (I'm PR bingo winner! GHO-98 on PR 98!)

orakili: @rupl, good catch, we should probably move that out of `gho-text` then because I'm not sure it's always loaded.

orakili: @rupl That said, thinking about that again, the style not applying outside of `Text` paragraph was on purpose. Other places share the sane ckeditor text format but according to the PDF prototype, this style is only to be used in `text` paragraphs.

rupl: ah shoot. Unfortunately I came back to this tab after lunch and proceeded with the changes but GitHub did not reflect your newest comments. I will make a quick follow-up and revert my commit.

PR #97, opened 2020-11-13
Develop main merge

rupl:
I got a notification that `develop` is out of date with `main` when trying to do routine deploy in #96 

This is a merge cleanup branch since just wholesale merging `main` into `develop` without looking at what would happen seemed irresponsible.


rupl:
alright, so nothing to commit except other merge commits. I'll just let it happen and hopefully it fixes #96

PR #98, opened 2020-11-13
[GHO-98] admin style for highlight

rupl:
Get admin styles working for highlights.


rupl:
I'm gonna yolo all this together and give it a shot on dev.

PR #101, opened 2020-11-14
[GHO-94] Bug fixes following permission review

orakili:
Ticket: GHO-94

This PR does 2 things:

1. It hides all the "about text formats" below formatted text fields because it's useless as we don't allow text format selection
2. It fixes an issue with the source field validation for further reading links, fixes the display of the link title when none is provided just to be thorough as this cannot happen because of the third change: only allow external links and make link title mandatory.

Note: regarding the last point (external link and title), those were the original settings before moving the further reading links to a paragraph type where I overlooked the original settings:

Screen Shot 2020-11-14 at 21 25 21

## Testing

`drush cim`, `drush cr` and edit a story and check that there is no "about text format" below the `footnotes` and `source` fields and add a further reading paragraph and test with internal link, missing title, missing source etc.

### DIFF
--- a/config/core.entity_form_display.node.article.default.yml
+++ b/config/core.entity_form_display.node.article.default.yml
@@ -14,6 +14,7 @@ dependencies:
     - field.field.node.article.field_thumbnail_image
     - node.type.article
   module:
+    - allowed_formats
     - double_field
     - inline_entity_form
     - layout_paragraphs
@@ -111,7 +112,10 @@ content:
     settings:
       rows: 5
       placeholder: ''
-    third_party_settings: {  }
+    third_party_settings:
+      allowed_formats:
+        hide_help: '1'
+        hide_guidelines: '1'
     type: text_textarea
     region: content
   field_thumbnail_image:
--- a/html/modules/custom/gho_fields/src/Plugin/Field/FieldFormatter/GhoFurtherReadingLinkFormatter.php
+++ b/html/modules/custom/gho_fields/src/Plugin/Field/FieldFormatter/GhoFurtherReadingLinkFormatter.php
@@ -37,7 +37,7 @@ class GhoFurtherReadingLinkFormatter extends FormatterBase {
       $element[$delta] = [
         '#theme' => 'gho_further_reading_link_formatter',
         '#url' => $url,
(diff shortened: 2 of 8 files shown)


html/modules/custom/gho_fields/src/Plugin/Field/FieldFormatter/GhoFurtherReadingLinkFormatter.php
```diff
@@ -37,7 +37,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       $element[$delta] = [
         '#theme' => 'gho_further_reading_link_formatter',
         '#url' => $url,
-        '#title' => $item->title ?? $url->toString(),
+        '#title' => $item->title ?: $url->toString(),
```
rupl, line 40: I had no idea you could write a ternary like this.

config/core.entity_form_display.node.article.default.yml
```diff
@@ -111,7 +112,10 @@ content:
     settings:
       rows: 5
       placeholder: ''
-    third_party_settings: {  }
+    third_party_settings:
+      allowed_formats:
+        hide_help: '1'
+        hide_guidelines: '1'
```
rupl, line 118: I see this configuration repeated in many places and it's gone almost everywhere, but I found another instance in the Section Index paragraph type. I will push a similar commit onto this branch before merging.

rupl (approved):

config/core.entity_form_display.node.article.default.yml
orakili: Good catch, Thanks!

html/modules/custom/gho_fields/src/Plugin/Field/FieldFormatter/GhoFurtherReadingLinkFormatter.php
orakili: I don't remember in which version of PHP it was added. A few new operators have been added (ex: `?:`, `??`, `<=>`) that are quite convenient but **not** **not* error prone as shown above :-s

Ref: https://www.php.net/manual/en/language.operators.comparison.php

PR #102, opened 2020-11-15
[GHO-103] Enable caption styling for blockquotes inside text paragraphs

orakili:
Ticket: GHO-103

This enables the caption styling for blockquotes inside text paragraphs.

Not sure about the margins because there is no example of pull quotes in any of the documents.

### DIFF
--- a/html/themes/custom/common_design_admin_subtheme/css/styles.css
+++ b/html/themes/custom/common_design_admin_subtheme/css/styles.css
@@ -33,9 +33,15 @@
   font-weight: 700;
   line-height: 2.25rem;
 }
+.paragraph--type--text blockquote {
+  margin: 1.5rem 0;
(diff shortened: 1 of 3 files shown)


html/themes/custom/common_design_admin_subtheme/css/styles.css
```diff
@@ -33,9 +33,15 @@
...
+.paragraph--type--text blockquote {
+  margin: 1.5rem 0;
+  font-size: 0.75rem;
+  line-height: 1.5rem;
+  color: #707070;
+}
 
 /* this is inside other stuff like Story */
-.field--type-text-long {
+.paragraph .field--type-text-long {
```
orakili, line 44: This is to avoid having the "summary field" in the form from being affected by this rule.

PR #106, opened 2020-11-16
[GHO-105] Adjust the separators to be full or half of the width of the reading width

orakili:
Ticket: GHO-105

This adjusts the separator to be full width or half width of the `reading width` (720px). I'm wondering if that's correct or if we want to be relative to the `content width` (904px).


rupl (approved):
It works perfect everywhere except IE and using browserstack IE11 devtools, I couldn't thoroughly inspect the pseudo-element to try and debug to produce a solution. Adding a hardcoded `margin-left: 180px;` didn't seem to work (if separator is half of 720px, and its margins are half of the leftover 360px)

I'll file a follow-up for IE11.

rupl:
image

PR #107, opened 2020-11-16
[GHO-106] Remove credits and allow text format to photo galleries

orakili:
Ticket: GHO-106

This changes the caption field to a formatted text field and removes the credits field and use the credits of the first image instead, and finally makes the location field optional, for the photo galleries.


rupl (approved):
Functionally it works pretty well, but we have a few adjustments we'd need to make in order to allow all Text Paragraph features to work within the caption. I'm unsure whether it's required at this time but making note just in case.

image

PR #109, opened 2020-11-16
GHO-48: CD header

rupl:
# GHO-48 and GHO-70

I finally whacked enough moles to put this in a PR. There are tons of weird fixes in here because, well, it's tough to override the CD as a subtheme. Basically anytime there's a variable defined, and we want to change that variable, you have to go and copy the entire hierarchy of files that produce the final CSS property.

I kept browserstack open the whole time, and was even developing primarily in IE11 for a couple of these fixes. So I feel pretty good about the end result even if the code is gross.

The most glaring Weird Thing™ is that the width of the nav items in the Header is affecting the toggle indicator arrow. I solved it by setting a width on the `<button>` elements themselves. Trust me, I was desperate to find another solution, especially one not so bound to content. But I marked it and we'll have to revisit a couple times as the nav gets translated. My apologies in advance.

One positive note: I figured out the multilingual logos all on my own and it seems to work flawlessly! So mark that one as a small drupal win for Chris.

There was minimal work done to the Footer. Just enough to make it presentable. There will be more coming in terms of footer work in a separate PR.


orakili (approved):
That looks good, thanks! There is an issue with the Spanish logo (iso2 code is ES not SP - I tend to do the same mistake quite often...) and there are some styling adjustments to do in RTL.

I've taken care of those in https://github.com/UN-OCHA/gho-site/pull/111

rupl:
ha! I copied those files directly from our spanish-speaking art directors and somehow never clicked spanish language 😁  I should have noticed porque estoy aprendiendo español ahora mismo.

The other three definitely work though.

PR #110, opened 2020-11-17
[OPS-7078] Docker image bump - newer tools and faster build.

cafuego:
Symfony flex will concurrently bulk pre-fetch all packages and then install them, which speeds up things nicely.

Though the new docker image is tagged as -develop, it is considered stable in Alpine. Let's sue it, so we can be sure and re-tag it as -stable before go-live.

Ideally, the build step will use `--no-dev` before go-live so we don't install such things as behat in the image :-)


orakili (approved):
Tested composer, created a new local image etc. Things are still working so looks good to me.

PR #111, opened 2020-11-17
[GHO-48] CD header follow up

orakili:
Ticket: GHO-48, GHO-84

This is a follow-up of https://github.com/UN-OCHA/gho-site/pull/109 that fixes the name of the Spanish logo, removes the 98px hack and adjust the styling in RTL.

**Before**

Screen Shot 2020-11-17 at 17 24 17

- GHO-84: Last menu (most left) popup is not attached to the button
- Menu text not aligned right
- double border left on first menu (most right)
- min-width of 140px due to flex-basis: 140px

**After**

Screen Shot 2020-11-17 at 17 23 03

Also the highlight was still red:

Screen Shot 2020-11-17 at 17 28 56


rupl (approved):
Looking at this it should have been obvious that needing to set `flex-basis` meant I had other flex props available to use. Thanks for fixing my brain fatigue 👍

PR #112, opened 2020-11-17
[GHO-120] Fix issue with related articles

orakili:
Ticket: GHO-120

Fix the undefined variable issue with related articles in Arabic.


rupl (approved):
:shipit:

PR #115, opened 2020-11-17
[GHO-122] Ensure the subtheme variables are used for the cd header/footer

orakili:
Ticket: GHO-122

This makes sure the variable overrides from the subtheme are applied to the base theme by doing the following:

1. import base theme cd-variables
2. import subtheme cd-variables
3. import everything from base theme except cd-variables
4. import everything from subtheme except cd-variables


html/themes/custom/common_design_subtheme/sass/styles.scss
```diff
@@ -1,11 +1,78 @@
...
+@import '../../../contrib/common_design/sass/base/_helpers';
+@import '../../../contrib/common_design/sass/base/_typography';
+
+
 //——————————————————————————————————————————————————————————————————————————————
 // Common Design
 //
 // Imports everything in the `cd` folder and ensures that the templates provided
 // in this theme for the Header/Footer will carry Common Design branding.
 //——————————————————————————————————————————————————————————————————————————————
```
rupl, line 51: the comment is now a lie.

html/themes/custom/common_design_subtheme/common_design_subtheme.info.yml
```diff
@@ -33,6 +33,12 @@ regions:
 
 # Override libraries.
 libraries-override:
+  # Don't load the common design styles.css as its rules are included in ours.
+  common_design/global-styling:
+    css:
+      theme:
+        css/styles.css: ''
+
```
rupl, line 41: ah yes, the D8 version of FOAD!

rupl (changes requested):
I noticed a couple issues while testing but will make a PR onto this one to fix them up so we don't create regressions.

rupl:
Another awesome side-effect is that source maps now work for CD! very cool.

PR #116, opened 2020-11-18
GHO-80: soft footer

rupl:
# GHO-80

Add all of our stuff to the Soft Footer. I surprised myself by figuring most of this out, but have one question that would potentially make translation simpler. I'll comment inline.

The styles feel like they got out of hand but all these separators in various axes on different breakpoints made for a really strange setup (also since two of the blocks act nested in one breakpoint and flat in the other).

## Testing

`drush cim &&  drush cr` and then look at the footer for the acknowledgements and FTS/HumInsight logos. I did very thorough testing while developing because I did not want to revisit this 😁


html/themes/custom/common_design_subtheme/common_design_subtheme.theme
```diff
@@ -21,6 +22,29 @@ function common_design_subtheme_theme() {
...
+        'description' => NULL,
+      ],
+      'template' => 'block--soft-footer-huminsight',
+    ],
+    'soft_footer_acknowledgements' => [
+      'variables' => [
+        'title' => NULL,
+        'description' => NULL,
+        // 'cerf' => Link::fromTextAndUrl(t('Central Emergency Response Fund'), Url::fromUri('https://cerf.un.org')),
+        // 'cbpf' => Link::fromTextAndUrl(t('Country-based Pooled Funds'), Url::fromUri('https://pfbi.unocha.org')),
```
rupl, line 44: I just could not figure out how to produce the markup for these two block variables, which would have simplified the translation by wrapping them in `{% trans %}` 

The `// use Drupal\Core\Link;` at the top also belongs to these two lines.

orakili (approved):
Looks great.

I removed the themes in `common_design_subtheme.theme` because they are not used (you would have to add a `'#theme' => 'soft_footer_acknowledgements'` to the block::build() for that).

There is a tiny issue in IE11 < 700px. I'll put the fix in another PR so you can confirm.

And another PR for a possible answer to your question regarding the CERF and CBPF links.

PR #117, opened 2020-11-18
GHO-48: colors and OCHA Services

rupl:
# GHO-48

Small follow-up. Emma noticed the arrows were wrong color and I fixed OCHA Services label while I was in there.


orakili (approved):
Thanks. I added a fix so that the "Menu" button's arrow also uses the UN blue color:

**before**

Screen Shot 2020-11-19 at 13 14 51

**after**

Screen Shot 2020-11-19 at 13 49 21

There is an issue with the "OCHA Services"  and the "user menu" on small screens when logged in but I don't think any of the people who can log in the site will do so on mobile, so probably not worthy to spend time on that:

Screen Shot 2020-11-19 at 13 10 49

rupl:
Nice catch and yeah that seems to be an internal issue which the public would never glimpse.

PR #119, opened 2020-11-19
[GHO-123] Fix bleeding

orakili:
Ticket: GHO-123, GHO-111, GHO-105

This PR is mainly to fix the bleeding by actually removing it and move the constraint to the elements that don't bleed.

The bleeding, with the negative margins didn't fully work in IE11 because, even with the ponyfill, updating the css variable that adjusts the margin based on scrollbar width had no effect.

### GHO-123 - Bleeding

So, the idea here is to do the opposite: instead of marking an element as bleeding, we mark elements that don't bleed as constrained within the `content-width`:

On node pages, the main content area's left and right paddings and max-width are removed (so it spans to 1400px) and by default all the children use the full-width. Then a `content-width` class is added to the elements that should be contained in the `content-width` (904px) area. This is handled via the `gho-page-node` "component".

The `sub-article`, `section index`, `story` and `facts and figures` and `more from this section` components have the `content-width` on their direct children in order to have the hero image or the top border use the full-width. All the other components have the `content-width` on the top element.

Additional templates were added to ensure we can add the `content-width` class to the appropriate components.

This approach has the advantage of not needing a script and adding the `content-width` is not much more complicated than adding the previous `gho-bleed` one:

You want something to use the full width, put the `content-width` on the children otherwise put the `content-width` on the parent.

**Bleeding black background**

The bleeding black background was simplified a lot. The `gho-aside--dark` has a dark background rule and the rest is controlled as described above by putting the `content-width` class on the children.

**Paddings**

The paddings that were on the `<main>` have been moved to the `content-width`. See `gho-page-node`.

### GHO-111 - Facts and figures and photo galleries

While working on this, I took the opportunity to fix the wrapping that occurs in IE11 by moving the width constraint in multi-column mode from the `flex-basis` to the `width` attribute so that IE11 respects the padding and negative margins of the parent.

### GHO-105 - Separators

There is also a simplification of the rules for the separators that has the bonus effect of fixing the issue in IE11 which didn't like the `:before` on a `<hr>` element.


html/themes/custom/common_design_subtheme/components/gho-section-index/gho-section-index.css
```diff
@@ -1,12 +1,11 @@
 .gho-section-index {
-  max-width: var(--content-width);
   margin: 3rem 0;
-  padding: 2rem 0 0 0;
+  padding: 3rem 0 0 0;
```
orakili, line 3: GHO-59: misread the value in the Gho-VisualAdjustments.pdf

rupl (approved):
This works well and seems to be consistent across all browsers. I like that we're ridding ourselves of some JS complexity too!

One thing that does occur to me having reviewed this: we're never changing the values of `--reading-width` or `--content-width` and they might as well be hardcoded. That means we don't need to run any JS at all to build the layout in the slowest JS engine we're targeting.

I attached a video to the ticket showing how even after CSS loads you get one more major shift before hitting Visually Complete threshold. I'm not suggesting that it be changed now because it's essentially a find/replace if we want to do it. And what we have here does end up looking great once JS loads.

orakili:
@rupl Yes, I think we can remove those css variables now that we have done most of the styling work (I think/hope). We need to check if there are other variables from the CD in use and then remove the ponyfill. 

That's not going to be super funny because it's a dependency of several libraries in the `common_design` base theme, so lots of overrides to do. 😢 I can take care of that.

rupl:
Let's just let it be for now though, right? No need to create work for ourselves. It's all functioning just fine aside from the layout shift.

orakili:
Yep, no urgency on that.

PR #120, opened 2020-11-19
[GHO-80] Fix wrapping issue of soft footer in IE11

orakili:
Ticket: GHO-80

This is a follow-up of https://github.com/UN-OCHA/gho-site/pull/116 that fixes a tiny issue in IE11:

**before**

Screen Shot 2020-11-19 at 12 28 14

**after**

Screen Shot 2020-11-19 at 12 52 14

I had the same [issue](https://github.com/UN-OCHA/gho-site/pull/119) for the photo galleries and facts and figures and using the `width` instead of the `flex-basis` worked in all browsers. 

**Note:** This also works with the `flex: 0 1 auto` short notation but I've taken the (bad?) habit of making the flex rules explicit after the issues we had with IE...


rupl:
Nice ok good to keep in mind, maybe I'll make an IE11 flex-box poster for my office!

rupl (approved):
In the other PR I said I tested thoroughly and it was true, but I was probably only bending the BS window to 850 or so. Thanks for the catch.

Regarding notation: I also prefer the flex shorthand and in most cases there is no consequence so use it IMO. But it's clearly less buggy to write them each individually, since IE 10/11 will respect unitless `flex-basis` and IE11 will respect `calc()` when flex-basis is declared as an individual prop.

PR #121, opened 2020-11-19
[GHO-80] CERF and CBPF links

orakili:
Ticket: GHO-80

@rupl This is a possible answer to your question in https://github.com/UN-OCHA/gho-site/pull/116 about (if I didn't misunderstand).

That puts the CERF and CBPF links outside of the acknowledgements string.

To do what you wanted with the `hook_theme`, you would have had to specify the use of that theme in the `Block::build()` function of the Acknowledgements block via `'#theme' => 'soft_footer_acknowledgments' and add the `#cbpd` and `#cerf` items to the render array it returns.

Having the `Link(...)` in the `hook_theme` would not have allowed proper translation because the theme definitions from the `hook_theme` are cached.


rupl (approved):
Beautiful! Here's what it looks in Translation UI which is so much cleaner to translate:

image

rupl:
@left23 if you look at this diff and the screenshot I dropped above to give you a QA goal, this is probably the most ideal method to translate sentences of text containing links.

PR #125, opened 2020-11-19
[GHO-111] ensure negative margins are enforced separately from content-width

rupl:
# GHO-111

Small fixup to get our styles (which were correct to begin with) applying cleanly within the `.content-width` container class.


orakili (approved):
Good catch, thanks!

PR #126, opened 2020-11-19
[GHO-80] set padding to match visual specs provided in design PDFs

rupl:
# GHO-80

small follow-up based on Gho-VisualAdjustments.pdf


orakili (approved):
I would have selected 20px for the padding so that 60 + 20 + 20 = the 100px from the document and so a bit less magic but eh.

Screen Shot 2020-11-19 at 23 45 20

PR #128, opened 2020-11-19
GHO-54: signature paragraph

rupl:
# GHO-54

~Image still pending but all the other stuff~ Everything including image is there ~and I figure if we get this on production and someone sees it, they'll prioritize getting us a final graphic more quickly.~

Margins for paragraph were picked by me. Not compared against anything since our v2/v3 PDFs show a very distorted graphic. If we come up with rationale or a better idea feel free to throw a commit on there and modify.

This was such a simple process I even threw some admin styles in there:

image


rupl:
Whoops forgot the themed screenshot 😁 

image

rupl:
Not bothering to update screenshots which had FPO PNG. SVG is in branch and much nicer-looking.

orakili (approved):
Thanks! I've adjusted the margins etc. to have an almost exact match with what is in the document (signature background on the screenshot not included obviously):

Screen Shot 2020-11-20 at 10 57 06

**Note**: The signature was not distorted when opening the PDF with an editor, so the dimensions are based on that.

PR #129, opened 2020-11-20
[GHO-127] Add styling for download link in main menu

orakili:
Ticket: GHO-127

Minimal styling to have the a dark blue download button in the main menu:

Screen Shot 2020-11-20 at 11 57 09

Screen Shot 2020-11-20 at 11 56 52

The behavior on hover or when active is unchanged because it's good enough in my opinion.

**Bonus**: They can add the download link where they want as the styling is not tied to the position of the menu element.


rupl (approved):
Works well enough and I merged #130 already so I'm hitting the button. I left some follow-up stuff in the ticket.

PR #130, opened 2020-11-20
[GHO-128] Hide tagline and download button on hero

orakili:
Ticket: GHO-128

This simply adds some `if` to only show the tag line and download button when the corresponding fields in the homepage form are filled in. This way if they change their mind, then we don't have to do anything (🤞).


rupl (approved):
Yup, good thinking. All four cases tested (both filled, summary, pdf, neither)

PR #131, opened 2020-11-20
[GHO-129] Upgrade to drupal 8.9.9 and update other modules

orakili:
Ticket: GHO-129

Drupal security update + other module updates.

Run `composer update` and `drush cr`


rupl:
Ok after some standard-for-me upgrade troubles it seems good. I didn't edit every last piece of content, but the Articles still work well, everything renders the same and so forth.

PR #133, opened 2020-11-20
[GHO-130] Replace non breaking spaces

orakili:
Ticket: GHO-130

This adds a text filter to remove the non-breaking spaces introduced when copying text from Word.


rupl (approved):
Standard chicken-and-egg error when setting up, but otherwise looking good after `cim && cr && cim`. I didn't paste from Word but manually adding spaces shows it's working:

image

PR #135, opened 2020-11-20
[GHO-48] color of menu buttons

rupl:
Title says it all ☝️


orakili (approved):
Looks good and I didn't find any other place using '#4a4a4a' in the GHO header.

PR #136, opened 2020-11-20
GHO-76: arabic font

rupl:
# GHO-76

Installing the Noto Kufi font, which will only be included on the page when rendering an Arabic-language URL. The code was copied from Google Fonts and added to our top-level twig template, in order to specify `font-display: swap` and avoid FOIT.

There are some other RTL-related bug fixes in there too. Seemed silly to let them be when I was checking everything.


orakili (approved):
That seems to be working as expected.

PR #137, opened 2020-11-20
[GHO-131][GHO-132][GHO-133] Homepage stuff

rupl:
# GHO-131 / GHO-132 / GHO-133

The Homepage is about to become a kitchen sink of weird exceptions. PR has the following:

- A Text paragraph using the Blockquote style will be the first paragraph on the homepage. The Homepage component treats it with special styles to center-align and unset `--reading-width`
- Next will appear the site logo but in a special format, (not the ones in the Header next to Nav) — I used another Text paragraph and entered an invisible character, so that more custom styling can position the image.
- Top-row figures are now centered.

There might be more if I can finish them today. I'll update the PR as I go.


html/themes/custom/common_design_subtheme/components/gho-home-page/gho-home-page.css
```diff
@@ -120,3 +124,73 @@
...
+ * Homepage customizations to existing paragraph types
+ *
+ * Some declarations are quite specific to ensure they never spill elsewhere and
+ * also override any selectors in the components themselves.
+ *----------------------------------------------------------------------------*/
+
+/**
+ * This represents the Caption that corresponds to the Homepage hero image.
+ */
+.node--view-mode-home-page .gho-home-page__content .paragraph--type--text:first-child {
```
emmajane, line 139: I'm being lazy: does the home-page requirement mean that I have to be on `/` or will these styles also work if I'm on a URL alias (is it even possible to navigate to this version of the homepage without it being on `/`)?

html/themes/custom/common_design_subtheme/components/gho-home-page/gho-home-page.css
```diff
@@ -120,3 +124,73 @@
...
+  margin-bottom: 0; /* see margin-top for logo */
+  text-align: center;
+}
+.node--view-mode-home-page .gho-home-page__content .paragraph--type--text:first-child .gho-text__text {
+  max-width: var(--content-width);
+  margin: 0 auto;
+}
+
+/**
+ * Homepage logo SVG
```
emmajane, line 152: we're okay to have an SVG here when we get them, right? it's just UGC that can't be SVG?

emmajane (approved):
🕹

html/themes/custom/common_design_subtheme/components/gho-home-page/gho-home-page.css
rupl: It's a view mode. we could use it elsewhere but I _hope_ we aren't.

I did review the other pages on the site to double check, for instance Top figures (aka Needs and Requirements paragraph type) do show up elsewhere and they are definitely NOT affected by this CSS.

html/themes/custom/common_design_subtheme/components/gho-home-page/gho-home-page.css
rupl: We'll see what he ends up giving us. If the SVG still contains font references (the first one did) then we'll have to use the PNGs he supplies.

If we sends us a fully vectorized(?) SVG that doesn't rely on fonts being present, then it can also be used as SVG background-image.

emmajane:
Thank you for responding to my comments. :shipit:

html/themes/custom/common_design_subtheme/components/gho-home-page/gho-home-page.css
rupl: Actually I take it back, the second class `.gho-home-page__content` is from a template that is used for Drupal's home page. We can be quite confident this stuff won't show elsewhere.

If somehow something leaks elsewhere, we can use `body.path-frontpage` or something drupal-specific for home page to clamp down harder.

PR #139, opened 2020-11-23
[GHO-131][GHO-132][GHO-133][GHO-135][GHO-136] Homepage

orakili:
Tickets: GHO-131, GHO-132, GHO-133, GHO-135, GHO-136

This implements the "new design" of the homepage in a more proper way. (Tested in IE, FF, Chrome, Safari, mobile etc.)

### GHO-131 Caption

This enables the article `caption` field for the homepage. It was hidden in the form and not rendered before. The credits of the hero image is injected at the end of the caption like it's done for normal articles.

### GHO-132 Logo

This adds the logo via a `img` tag in the content of the homepage so that it can adapt to the screen size.

It uses the SVG from the dropbox link. This SVG doesn't contain any reference to any font. Maybe the file was changed this weekend.

I was a bit hesitant in the approach: image tag in the content, background-image, image in the `<h1>` with `order` to change its position. In the end, I went with the image tag in the content as it was better for accessibility and styling.

### GHO-133 Needs and requirements

Nothing much to say there. It's centered, all the elements have the same padding so the middle figure is well aligned with the logo etc.

There was a bug not related to the homepage where the top right (visually) element in RTL was missing its left border.

### GHO-135 Intro

Intro paragraph below the logo with styling from the PDF (21px font-size etc.). I added some media queries to adjust the font-size otherwise it's way too large on mobile.

### GHO-136 Section index

Styling (font, margin) for the section index paragraphs has been adjusted. It's only used on the homepage so the changes were put in the `gho-section-index`.

This also adds a `field_paragraphs` to the section index that can accept photo galleries, interactive content and text paragraphs, just in case.

The image + caption shown in the PDF is a photo gallery with 1 image.

### Specificity

There is no need for the `.node--view-mode-home-page` class etc. because the `home_page` view more is only applied to the home page and the `gho-home-page` component only used for nodes using this view mode.

However to avoid affecting any potential sub-articles that could be included on the homepage, the special rules target the direct children of the `gho-home-page__content` element.

## Translations

There is one new string to translate: `Global Humanitarian Overview 2021 - Large logo` used for the `alt` of the logo image.

## Testing 

`Drush cim`, `drush cr` and add text to the homepage caption field, an interactive content as first paragraph, a text paragraph after that and a photo gallery to the first section index.

## Screenshots
gho homepage desktop
gho homepage mobile


html/themes/custom/common_design_subtheme/templates/nodes/node--article--home-page.html.twig
```diff
@@ -123,12 +124,22 @@
...
   #}
   {% if node.isPromoted() %}
-  <div{{ content_attributes.addClass('node__content').addClass('gho-home-page__content') }}>
+  <div{{ content_attributes.addClass('node__content').addClass('gho-home-page__content').addClass('content-width') }}>
+    {{ content.field_caption }}
+
+    {#
+      @todo review alt tag and/or hiding that from the accessibility tree as
+      there is already the title above in the header.
+    #}
```
rupl, line 133: This TODO is why I made it a bg-image originally. I didn't consider it necessary to declare a largely-decorative title in the content flow.

rupl (approved):
Everything looks good and yeah the homepage is a little less weird to content editors AKA us.

I saw a couple margins were off from PDF but, but the computed values are correct when viewing DOM. We can revisit if it becomes an issue. I'm fine with it.

image

PR #140, opened 2020-11-23
[GHO-109] Add the scroll indication at the bottom of the hero image

orakili:
Ticket: GHO-109

This adds the "scrolldown to explore" indication at the bottom of the hero image on medium/large screens (it takes too much space otherwise on mobile):

Screen Shot 2020-11-23 at 13 35 26

The SVG for the mouse was extracted from the  PDF v3 and cleaned/minimized via SVGO.

The font styling, positioning etc. was done by doing a direct comparison with the PDF v3.

## Translations

There is a new string to translate: `Scroll down to explore`.

## Testing

`Drush cr`


rupl (approved):
Looks good and if we get an official SVG I would make it more useful by animating the arrow or something. I posted a sample in the ticket.

PR #141, opened 2020-11-23
[GHO-126][GHO-139] Datawrapper embeds

orakili:
Tickets: GHO-126, GHO-139

This adds the possibility to embed datawrapper content.

This is done via a "embed code" text field in the `interactive content` paragraph type where one can drop the (responsive) `<iframe>` code from datawrapper which is then parsed and validated so that we don't blindly execute something else.

The field is not mandatory and the image field has also been made non mandatory.

Access to the embed code field is controlled by a permission so we can text embeds in dev before allowing admins to use it as per request in GHO-126.

This works pretty well and here are the screenshots of the homepage with the embedded table (GHO-139) on desktop and mobile:

Screen Shot 2020-11-23 at 13 52 15
Screen Shot 2020-11-23 at 13 52 27

## Testing

`drush cim`, `drush cr` maybe a couple times if drupal is not happy as usual. Then if you're using a normal admin account, check you don't see the `embed code` field in the form, then give the `add datawrapper embed code` permission to the `administrator` role and embed something (see GHO-139 for the embed code).


rupl (approved):
I didn't spend too much time trying to stress-test the actual input. I just copied the `<iframe>` code and tested how the website produces the embed, which worked very well. 

Nice job on the message handler to ensure it's always the right height. I filtered the data and it worked across all browsers/viewports. Even works on Alcatel 1X (low-powered CPU running modern Android Go).

PR #143, opened 2020-11-23
[GHO-138] Fix needs and requirements import

orakili:
Ticket: GHO-138

Well, working late at night is not a good thing... nor is not having automated tests...

This fixes an issue that was preventing importing new figures.

## Testing

`drush cr` and then import the file mentioned in the ticket.


rupl (approved):
Works as advertised! :shipit:

PR #145, opened 2020-11-23
[GHO-125] Achievements

orakili:
Ticket: GHO-125

This adds an "achievement" content type with a title, icon, sources and list of paragraphs (text, photos, interactive, separator and bottom figures).

It also adds an "achievement" paragraph type to select an "achievement" node similar to stories or sub-articles.

Screen Shot 2020-11-23 at 20 52 47

### Figures

This adds a `figures` formatter to be able to style bottom figures as "top figures". I wish we has used a generic "figures" for that paragraph type and associated css component but we didn't so this bridges the gap.

In the "bottom figure row" paragraph form, there is the "behavior" at the top that allows to select the view mode: "default" (bottom figures style) or "large figures" to apply the same style as the "Needs and requirements".

Screen Shot 2020-11-23 at 21 07 23

Screen Shot 2020-11-23 at 21 39 13

In an ideal world, I would rewrite the "Needs and reuquirements" to use the "gho-figures" classes but reality being what it is, instead, I added the `needs-and-requirements` classes to the `gho-figures-formatter--large.html.twig`.

### Icons

I dropped the "Humanitarian Icons" set into the `gho-achievement` component and the selection of the icon is done by entering the basename of the icon file in the `field_icon` on the achievement node form (ex: Food-Security).

### Styling

Aside of the figures, there is basic styling with probably "wrong" margins... but it looks good enough for now (see screenshot).

**Note** The individual achievement page are not styled (like the stories and sub-articles), that will be done in another ticket.

## Testing

`drush cim`, `drush cr`, possibly a few times, then create achievements via the `/admin/content/achievements` page then add them to an article.

## Screenshots

achievements


orakili:
@rupl The "421" changed files are because of the humanitarian icons. You can filter out `svg` files when viewing the changes, they are only from the set.

rupl (approved):
Flawless victory 💯

PR #146, opened 2020-11-24
[GHO-142] Main navigation too long - fix - dirty option 1

orakili:
Ticket: GHO-142

With the download button in the main navigation, it's too long and it starts overlapping with the GHO logo at around 1062px.

One option is to change the media query to show the "menu" button earlier.

Another option is to have the menu items (buttons + download link) shrink to use the minimum content width: be "content-aware" as it was described. This was much more complicated than it should have been:

**Be "content-aware"**

- ~~Simple `display: table-caption` on `span` inside button/link: works well except on IE11 (span has "no" width in that case)...~~

- There a nice `width: min-content` property that works well in most browsers but of course it's not supported by IE11. There is hack for IE11 by putting `display: -ms-grid` and `-ms-grid-columns: min-content` on the parent. That works fine enough.

- To use `min-content` we need to remove the `display: flex` on the buttons/links and use `display: block`: we lose the vertical centering for the links: fine enough, we can override the template and add a span inside the `a` and center it.

- In addition to the above, we need to position the `arrow` absolutely, easy enough with a padding right (or left in RTL) on the button.

**Breaking the text**

- We don't want to have `Inter-agency appeals` break on the hyphen: fine, we can use `word-break: keep-all`, right? Nope, that doesn't prevent IE11 from breaking on the hyphen... None of the other `word-wrap` etc. properties can do that either. So dirty solution: replace the hyphen with a [non-breaking hyphen](https://en.wikipedia.org/wiki/Wikipedia:Non-breaking_hyphen) - works fine...

- We don't want the `Introduction and foreword` to break on every space:  no choice on the css side, so add a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space#Keyboard_entry_methods) between the `and` and `foreword`.

Strings with the breaking spaces/hyphens for easy copy and paste:

1. `Introduction and foreword` 
2. `Inter‑agency appeals`

**Conclusion**

Well that works even though I'm not super happy with having to use non-breaking spaces and hyphens. Result:

**LTR**
Screen Shot 2020-11-24 at 15 59 31

**RTL**

Screen Shot 2020-11-24 at 16 29 27

**At 1025px before the switch to the "menu" button**

Screen Shot 2020-11-24 at 16 06 12


orakili:
Closing in favor of https://github.com/UN-OCHA/gho-site/pull/147

PR #147, opened 2020-11-24
[GHO-142] Main navigation too long - fix - dirty option 2

orakili:
Ticket: GHO-142

Another option for GHO-142 that is a bit less hacky than https://github.com/UN-OCHA/gho-site/pull/146

It changes the field type for the menu link titles to a textarea so we can add line breaks to control where to break the menu labels and use `white-space: pre` to respect that.

This removes the need for the `-ms-grid`, `min-content` etc.


html/themes/custom/common_design_subtheme/sass/cd/cd-header/_cd-nav.scss
```diff
@@ -117,6 +122,15 @@
             flex-shrink: 0;
           }
         }
+
+        // While buttons keep their content centered vertically in display:block
+        // links don't so we need to center its content another way.
```
rupl, line 127: yay 😄

rupl (approved):
I think for the situation it's quite a nice solution. Re-tested everywhere and it looks the same but with no logo overlap :shipit:

PR #150, opened 2020-11-24
[GHO-143] convert CD Footer back to proper color

rupl:
# GHO-143

I must have misunderstood last week. I made the footer text black. This restores both CD Header/Footer with `#f1f1f1` and `#1a344a` as the branding colors.


orakili (approved):
Looks good to me.

PR #151, opened 2020-11-24
[GHO-150] adjust Appeals tags

rupl:
# GHO-150

Winner of tiniest PR contest?


orakili (approved):
Looks good to me. I think that's the only place where we still are using `em` instead of `rem` for info.

rupl:
You know, I intentionally used `em` units because I mistakenly thought that the appeals tags were wrapped inside the `<h2>` for the article title, and wanted them to stay proportional to the parent font-size. I see now by reviewing the DOM that it's not the case, so using `em` is an anomaly.

PR #153, opened 2020-11-25
[GHO-121] Always use dark grey for highlight style

orakili:
Ticket: GHO-121

Another short PR. 

**Note**: This results in having blockquotes with the highlight style applied to look like other paragraphs with this style, so let's wait for confirmation before merging.


rupl (approved):
Amazing work, as always.

PR #155, opened 2020-11-25
[GHO-126] Fix datawrapper iframe ID

orakili:
Ticket: GHO-126

Minor change to restore the ID prefix for the datawrapper iframes. That doesn't have any impact but it's cleaner.


rupl (approved):
Agreed, always good to namespace especially with the ID being supplied by a 3rd party.

PR #156, opened 2020-11-25
[GHO-69] update generic hyperlink color

rupl:
# GHO-69

We have a new shade of blue!


orakili (approved):
Checked all the places with links I could think of and the new color seems to be used.

PR #157, opened 2020-11-26
GHO-155: achievement icon

rupl:
# GHO-155

Moving Achievement icons over to the end of reading-width column, and aligning them vertically with top edge.


orakili:
Didn't mean to close it... Stupid layout shift.

html/themes/custom/common_design_subtheme/components/gho-achievement/gho-achievement.css
```diff
@@ -7,32 +7,36 @@
   display: flex;
   flex-wrap: nowrap;
   flex-flow: row;
-  align-items: center;
+  align-items: flex-start;
+  max-width: var(--reading-width);
```
orakili, line 11: This is ignored because the header has the `content-width` class in the templates. I think that's fine because this way it's aligned with the photo gallery below and looks better:

**Content-width**

Screen Shot 2020-11-27 at 12 39 47

**Reading-width**

Screen Shot 2020-11-27 at 12 39 10

Above done via putting `max-width: calc(var(--reading-width) - 2rem - 60px);` on the `.gho-achievement__heading`.

orakili (approved):
Looks fine.

html/themes/custom/common_design_subtheme/components/gho-achievement/gho-achievement.css
rupl: Hm good idea to test with a gallery. I used a page a Text paragraph first as my test content which is why I picked `reading-width`.

PR #158, opened 2020-11-26
GHO-152: datawrapper updates

rupl:
# GHO-152

1. Adds a checkbox to the Interactive form so that we can decide on an individual basis whether the fallback image or the Datawrapper embed displays on the website.
2. Adds permission to `administrator` to alter Datawrapper embed fields. Edgar can now edit or update embeds. This will be useful as he gets around to enabling RWD mode on the embeds, and won't have to ask us in order to update.

Standard testing stuff: `drush cim` and `drush cr`, then find an Interactive paragraph and ensure that it has both image/embed filled out. Toggle checkbox and observe result on the site.


orakili (approved):
Does what it says. I added a commit to restore the display of the other fields (currently only the `source` field as per discussion in flowdock and moved the checkbox closer to the datawrapper field because I'm afraid it would be overlooked otherwise.

rupl:
I didn't realize there were more potential fields to output. Thanks for that!

PR #159, opened 2020-11-26
[GHO-157] fix clipping of bottom row figures

rupl:
# GHO-157

Seems like our negative margin combined with special treatment of first-child was resulting in minor clipping.


orakili (approved):
Looks good. I added some fixes on RTL:

**Before**

Screen Shot 2020-11-27 at 13 13 43

**After**

Screen Shot 2020-11-27 at 13 17 15

rupl:
Thanks. Those changes aren't actually necessary on real content, because there are at least two `dir="ltr"` attributes when testing via DOM manipulation: one on `html` and one on `article` itself.

To remedy, I'm trying to get in the habit of prefixing LTR/RTL  stuff with the tag itself instead of just the attribute:

```css
html[dir='rtl'] .class {
  /* certain to work as intended! */
}
```

rupl:
That kind of prefixing would be best stored in a Sass mixin IMO so we can globally control it, but aside from minor conveniences like that, I have really enjoyed using plain CSS on these components.

orakili:
I agree and in theory we are not going to show mixed content, especially Arabic articles inside an English page but I thought it would be more robust to make all the margins/padding/border explicit to avoid any surprise, notably when having deeply nested nodes.

PR #160, opened 2020-11-26
[OPS-7129] Make config_split config be the same as IASC.

cafuego:


orakili (approved):
Approving because we indeed don't want those dev modules on prod.

However the problem with this approach is that it will uninstall `config_split` when deploying to prod (which is fine, we don't need it there) but then when importing a prod DB dump in a local environment everything in `config-dev` will be ignored.

In that case, one has first to re-enable all the modules from `config-dev/config_split.config_split.development.yml`:

1. `drush pm-enable config_split dblog devel webprofiler`
2. `drush cim --partial --source=/srv/www/config-dev`
3. `drush cim`

cc @rupl

PR #162, opened 2020-11-27
[GHO-153] Fix footnotes

orakili:
Ticket: GHO-153

Due to Drupal's render caching, the hook_preprocess that were called to accumulated the footnotes and show the references list at the bottom of the page was not working on second visit.

So this PR deals with that differently using a post render callback which is called after generating the HTML to extract the texts and footnotes from it and generate the list. 

Because this a post render callback on the node, the result gets cached at the page level so there is no impact for anonymous users.

This also adds the missing `preview` view mode for the achievement nodes and paragraphs so they can be rendered wit the references at the bottom of each paragraph in the form.

## Testing

drush cr, drush cim, then visit a page with paragraphs containing footnotes and ensure they are accumulated at the bottom of the page


rupl:
As a functional review, it does seem to accumulate everything at the bottom. I tested by numbering all the footnotes according to their paragraphs and they all come out at the bottom in the right order, with incremental numbering across the URL instead of paragraph:

image

image

rupl (approved):

PR #165, opened 2020-11-27
[GHO-160] get 3RP appeals tag to be correct color

rupl:
# GHO-160

Drupal trying to be a little too careful here..


orakili (approved):
Working as expected.

PR #170, opened 2020-11-30
[GHO-159] add HTML ID to sub-article Titles

rupl:
# GHO-159

Add HTML ID into markup for all Sub-Articles. Simple change to Twig.


orakili (approved):
That works well. I was afraid `clean_class` would make a mess with Arabic titles for example but no problem (not test in IE though).

rupl:
It works in IE6+ per our flowdock discussion and this reference: https://mathiasbynens.be/notes/html5-id-class

PR #171, opened 2020-11-30
[GHO-167] Homepage v3

orakili:
Ticket: GHO-167

This adds an "article list" paragraph type for the links in a section index with an optional title and a list of internal links. There is a new formatter to be able to get the node titles for the links.

Section index paragraph types are modified with an image field, a caption (location + text) like articles and can now only have "article list" paragraphs.

Styling is changed to match what is in the new PDF.

**Note**: the margins are too big in my opinion on smaller screens but no values were provided so....

## Testing

`Drush cim`, `drush cr` (possible a few times) then edit the homepage, and fill the section index paragraphs.

### DIFF
--- a/config/core.entity_form_display.paragraph.article_list.default.yml
+++ b/config/core.entity_form_display.paragraph.article_list.default.yml
@@ -1,0 +1,34 @@
+uuid: 1c3a01a2-6468-4567-bb77-f2d64d8121eb
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.paragraph.article_list.field_links
+    - field.field.paragraph.article_list.field_title
+    - paragraphs.paragraphs_type.article_list
+  module:
+    - link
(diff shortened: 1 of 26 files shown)


rupl (approved):
Overall looking really slick!

So far the only thing that seems off to me was the Introduction's title. Until I filled in the Section field, (for specifying Part zero, Part one, etc) the title rendered white-on-white below the graphic. Once I added something it popped into place:

image

image

I don't expect that content to change so often, but just that we're aware of why that happened. Other ones rendered just fine from the start since they had the field filled in:

image

In terms of testing, I gave this a go on the following:

- macOS Chrome/FF/Safari
- iOS Safari (iPhone 11)
- Win7 FF/IE11
- Win10 Edge 87
- Android GO 8.1 Chrome
- Android standard 8.1 Chrome/FF/Brave

PR #172, opened 2020-11-30
[GHO-151] Display next article instead of more from this section

orakili:
Ticket: GHO-151

This replaces the logic to get the related articles to get the next accessible article in the order of the main navigation. So for example, for the last article of the "Introduction and Foreword" menu, the link will be the first article of the "Trends and challenges" menu etc.

There is no change to the styling.

## Translations

`Next article`

## Testing

`drush cr` and check **as an anonymous user** that navigation via the "next article" is the same as the main navigation.


rupl (approved):
Tested all of it as anon and works as advertised :shipit:

PR #177, opened 2020-12-01
Main develop conflict resolution 20201201

orakili:
Attempt to make sure that we can merge `develop` into `main` in the future.


rupl (approved):
Ok, fully tested one more time:

- GHO-151: Homepage doesn't crash AND the individual articles now show "Next Article" instead of the entire menu index for that section.
- GHO-167: Homepage design is updated and functions as intended.
- GHO-169: OCHA Services still same as hotfixed `main`
- GHO-159: HTML anchors still same as hotfixed `main`

PR #179, opened 2020-12-02
[GHO-170] Override the media oembed resource fetcher to handle youtube quirks

orakili:
Ticket: GHO-170

It happens that the issue with the youtube videos is due to the youtube oembed proxy that doesn't always return the correct content-type header: `application/json` and sometimes wrongly `text/html` while the response body is a json string with the embed data.

So this overrides the media oembed resource fetcher service to deal with that quirk.

## Testing

`drush cr`.

@rupl This cannot be really tested aside of making sure the videos still appear, so you'll have to trust me on this and just check I didn't typo-ed anything in the code ;-)

### DIFF
--- a/html/modules/custom/gho_general/src/OEmbed/GhoResourceFetcher.php
+++ b/html/modules/custom/gho_general/src/OEmbed/GhoResourceFetcher.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Drupal\gho_general\OEmbed;
+
+use Drupal\Component\Serialization\Json;
+use GuzzleHttp\Exception\RequestException;
+use Drupal\media\OEmbed\ResourceFetcher;
+
+/**
+ * Fetches and caches oEmbed resources.
+ */
+class GhoResourceFetcher extends ResourceFetcher {
+
+  /**
+   * {@inheritdoc}
+   *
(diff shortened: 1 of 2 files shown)


html/modules/custom/gho_general/src/OEmbed/GhoResourceFetcher.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+    elseif (strstr($format, 'text/javascript') || strstr($format, 'application/json')) {
+      $data = Json::decode($content);
+    }
+    // Sometimes the youtube oembed proxy returns the wrong content-type while
+    // the body is the correct JSON.
+    elseif (strstr($format, 'text/html') && strpos($url, 'youtube') !== FALSE) {
+      $data = Json::decode($content);
+    }
+    // If the response is neither XML nor JSON, we are in bat country.
```
cafuego, line 50: 🦇

orakili: That's from the original `Drupal\media\OEmbed\ResourceFetcher::fetchResource()`. Not sure what they have with 🦇 .

rupl: bat-country

rupl (approved):
I see videos on my local.

Can't stop here!

PR #180, opened 2020-12-02
[GHO-175] Homepage adjustments

orakili:
Ticket: GHO-175

Minor styling fixes for the homepage v3


html/themes/custom/common_design_subtheme/components/gho-section-index/gho-section-index.css
```diff
@@ -24,30 +24,35 @@
...
+    line-height: 2.65rem;
   }
 }
 @media (min-width: 1024px) {
   .gho-section-index__title {
     font-size: 3.75rem;
-    line-height: 3.25rem;
+    line-height: 3.75rem;
   }
 }
```
rupl, line 57: I think `.gho-section-index__pre-title` needs to be set in the 1024px MQ instead of 567px. I see in the ticket:

> Small title on top of picture – increase letter-spacing to 0.15rem
> ...
> On tablet/mobile, please edit as follows:
> ...
> Small title on top of picture – decrease size to 0.675rem

So the font-size should be .675rem until "desktop"

I tried to use the GitHub suggestion to make it easier, but it doesn't work on deleted lines :(

rupl (approved):
Approved with one change.

html/themes/custom/common_design_subtheme/components/gho-section-index/gho-section-index.css
orakili: Thanks, I added commit to change that.

PR #181, opened 2020-12-02
[GHO-176] Adjust display of soft footer

orakili:
Ticket: GHO-176

This removes the desktop media queries for the soft footer so that it has the same appearance regardless of the screen size:

Screen Shot 2020-12-02 at 15 54 11


rupl (changes requested):
I spent toooo much time making this design work in two drastically different ways.......... sure let's throw it out.

Next we'll need an icon to instruct users "scroll down to see past the soft footer" 🤣 

I'm happy to handle this, but one of the borders is no longer "full bleed" and I see it in Chrome/Safari/FF on mac. It worked before, because at 1024px we broke out into the new layout. I re-colored the element red to make it easier to see:

image

orakili:
Thanks, nice catch. I added [commit](https://github.com/UN-OCHA/gho-site/pull/181/commits/8f45d9d4cd07759cdbda3254b6a0e93678ecd42f) to fix that so that this border always bleed to the edges (the overflow on the soft-footer limits it to correct area).

rupl:
💯

PR #182, opened 2020-12-02
[GHO-86-87-88] Improve admin content views

orakili:
Tickets: GHO-74, GHO-86, GHO-87, GHO-88

This improves the views used for the `/admin/content/` pages:

- Add translation overview for media and add `videos` and `authors` pages
- Add `destination` parameter to go back to the overview after editing or creating a node, term or media
- Remove links to term pages because they don't make sense for the GHO site
- Link the Needs and requirements to their edit page

Screen Shot 2020-12-02 at 21 33 50

Screen Shot 2020-12-02 at 21 32 21

Screen Shot 2020-12-02 at 21 35 46

## Testing

`drush cim`, `drush cr`, go to the different `/admin/content` pages, check they all show an overview (NaR is a bit different with the actual figures) and click on some `edit` and `create` links

### DIFF
--- a/config/views.view.content.yml
+++ b/config/views.view.content.yml
@@ -67,18 +67,18 @@ display:
           summary: ''
           description: ''
           columns:
-            node_bulk_form: node_bulk_form
+            nid: nid
             title: title
             type: type
-            name: name
-            status: status
             changed: changed
-            edit_node: edit_node
-            delete_node: delete_node
-            dropbutton: dropbutton
-            timestamp: title
+            view: view
+            view_1: view_1
+            view_2: view_2
+            view_3: view_3
           info:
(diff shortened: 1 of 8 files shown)


emmajane:
Just to confirm: does this get rid of the sort by date option? I'm not sure what the second screenshot is showing...

I'm asking because the sort by date option is actually really useful to me and it's been annoying on more than a few occasions to not have it on /admin/content. I'm not sure at what point it was removed (maybe adding translations removes it to save space). I would be against removing it from any further screens.

orakili:
Fair point, the "changed" column has been added to all the `/admin/content` pages.

rupl (approved):
LGTM :shipit:

PR #183, opened 2020-12-02
[GHO-7][GHO-173] social buttons + appeals tags

rupl:
# GHO-7 + GHO-173

Adds social links to Sub-Articles and Story paragraphs. The links don't show for top-level Article pages (the pages in the main Navigation).

This is still waiting on final copy from Edgar/Lillian so it's marked draft.

### Outstanding issues

The Section/Article titles take up lots of room and the Article titles frequently wrap, meaning we can’t just calculate a vertical position from the top edge, making some room by way of padding the nearby elements.

The Hero and its caption mean we can’t calculate from the bottom of the heading area, since the caption is of unknown height.

I settled on a `<footer>` element at the bottom (source-order wise) of each Article/Sub-Article/Story, which uses flex-box to change visual order to show it at the top. There’s a ton of picky styles holding it in place and I still can’t nail them in FF, which seems to select a different image than Chromiums/Safari.

I adjusted the aspect ratio of that image to always be one ratio. That seems to fix my FF problem in my local.

### Translation

Added to GSS 👍 

- `Share on Twitter`
- `Share on Facebook`
- `Share on LinkedIn`
- `Copy URL to clipboard`
- `URL copied to clipboard`

## Testing

`drush cim`, `drush cr` and load most any article besides the homepage and look for the social buttons in the upper-right, or left on RTL. The article for RTL must have a translated Media entity AND don't forget to translate the Section Title (e.g. "Part Two: Inter-Agency Coordinated Appeals") — without both the layout looks wrong.

The three social networks should link externally, and the last button should tell you it copied the URL of the content to the clipboard.


rupl:
We're also waiting on final icons. Sorry for the premature review request.

html/themes/custom/common_design_subtheme/common_design_subtheme.theme
```diff
@@ -158,6 +158,17 @@ function common_design_subtheme_preprocess_node(&$variables) {
...
+  // Add the SVGs for gho-social-links so we can output them in multiple templates
+  // without having to keep them up to date.
+  $icon_twitter = file_get_contents(drupal_get_path('theme', 'common_design_subtheme') . '/components/gho-social-links/img/icon--twitter.svg');
+  $icon_facebook = file_get_contents(drupal_get_path('theme', 'common_design_subtheme') . '/components/gho-social-links/img/icon--facebook.svg');
+  $icon_linkedin = file_get_contents(drupal_get_path('theme', 'common_design_subtheme') . '/components/gho-social-links/img/icon--linkedin.svg');
+  $icon_copyurl = file_get_contents(drupal_get_path('theme', 'common_design_subtheme') . '/components/gho-social-links/img/icon--copyurl.svg');
+  $variables['icon_twitter'] = $icon_twitter;
+  $variables['icon_facebook'] = $icon_facebook;
+  $variables['icon_linkedin'] = $icon_linkedin;
+  $variables['icon_copyurl'] = $icon_copyurl;
```
orakili, line 171: Need the PDF download icon (for GHO-29).

html/themes/custom/common_design_subtheme/components/gho-article/gho-article.css
```diff
@@ -48,14 +59,21 @@
...
   padding: 0;
-  margin: 0 0 3rem 0;
+  margin: 0;
   border: none;
   font-size: 2.25rem;
   line-height: 2.375rem;
   color: #1f1f1f;
 }
+html[dir='ltr'] .gho-article .page-title {
+  padding: 0 94px 0 0; /* make room for social links */
```
orakili, line 72: Need more space for the download PDF icon (for GHO-29).

html/themes/custom/common_design_subtheme/common_design_subtheme.theme
```diff
@@ -158,6 +158,17 @@ function common_design_subtheme_preprocess_node(&$variables) {
     $variables['title'] = common_design_subtheme_get_block_render_array('page_title_block');
     $variables['local_tasks'] = common_design_subtheme_get_block_render_array('local_tasks_block', FALSE);
   }
+
+  // Add the SVGs for gho-social-links so we can output them in multiple templates
+  // without having to keep them up to date.
+  $icon_twitter = file_get_contents(drupal_get_path('theme', 'common_design_subtheme') . '/components/gho-social-links/img/icon--twitter.svg');
```
orakili, line 164: Rather than getting the content here, it might be better to include the content in the twig template via `{{ source(directory ~ '/img/icon--twitter.svg') }}` etc.

html/themes/custom/common_design_subtheme/templates/nodes/node--story--teaser.html.twig
```diff
@@ -105,4 +106,7 @@
     </div>
     {{ content|without(['field_type', 'field_media', 'field_text', 'credits']) }}
   </div>
+
+  {% set ghoSocialLinksClass = 'gho-story__social-links' %}
+  {% include '@common_design_subtheme/content/gho-social-links.html.twig' %}
```
orakili, line 111: For info, you can pass the variable via `{% include '@common_design_subtheme/content/gho-social-links.html.twig' with {'ghoSocialLinksClass': 'gho-story__social-links'} %}`

html/themes/custom/common_design_subtheme/common_design_subtheme.theme
rupl: at some point I attempted that exact approach, but I was never able to output the file 😢

html/themes/custom/common_design_subtheme/templates/nodes/node--story--teaser.html.twig
rupl: Thanks, that's good to know. It feels less like polluting globals when passing directly.

rupl:
@orakili ok it's finally ready. all the icons were swapped, I adopted your method of sending args which is very handy to know, and I re-tested in Chrome/FF/IE11 for all buttons including copy-to-clipboard.

html/themes/custom/common_design_subtheme/components/gho-article/gho-article.css
rupl: thanks for the reminder. new ✨  magic number ✨  in place 😂

html/themes/custom/common_design_subtheme/common_design_subtheme.theme
rupl: 👍

rupl:
Oh the PDF button is also in there, but the HTML is commented out. All CSS should be adjusted to accommodate it when it appears, including its unique width of 19px.

orakili (approved):
Working well.

PR #186, opened 2020-12-03
[GHO-177] Multilingual

orakili:
Ticket: GHO-91, GHO-177

This PR enables adding content in translated nodes. Currently it's not possible to edit paragraphs inside a translation. This solves the problem.

There are lots of changed files in this PR but it's mostly configuration to enable translation of paragraphs.

### GHO-177: enable translations

In addition, this module adds the "published" checkbox for paragraphs to allow aymetric translations: the English version is the main with the full content and in translations, it's possible to "unpublish" paragraphs for which there is no translation available for example.

This also adds a visual help to flag `unpublished` and `untranslated` content:
- dashed orange outline for unpublished
- solid orange outline for untranslated
- solid red outline for untranslated and unpublished

This replaces the previous red/green banner that was indicating the language + status of nodes.

This also enabled contextual links (tiny icon on top right of media for example) to display a menu to jump to the edit or translation page for the media. Combined with the visual help, it makes it easier to create missing translations.

### GHO-91: align form

This makes the Arabic form aligned right which, even if it's sometimes awkward with English text, should help entering Arabic content.

### Styling fixes

For some reason on translated page, the `gho-caption` component on the homepage was loaded after the `gho-page-node` and a rule was overriding the left and right margin from the `content-width` class. This fixes it.

Also, this changes the `article list` template to hide the list title if it has no children.

## Testing

`drush cim`, `drush cr` and try to translate some articles, media etc.


rupl (commented):
For starters: I did all of the following as user/3 with Administrator privs. Absolutely no testing was done with user/1 😁 

Everything with a checkbox I was indeed able to accomplish, even if some of the notes explain potential issues I believe I uncovered. However it's a complicated system and I might still be confused about the novel way of highlighting translation issues. I'm hoping if I point something out and somehow misunderstood, we can turn our discussion into some editing docs for Emma and co.

- [x] Editing an existing translation not created by my user - was able to edit things made by both Admin, edgarpinho, and chrisruppel.
- [x] Translating Media fields — showed orange border until translating.
- [x] Translating an existing EN paragraph and publishing
  - [x] Text 👍 
  - [x] Bottom row figures 👍 
  - [x] Facts and Figures — this seems to be reversed in the editor+preview. For instance I translated the first/third image, only then did the orange box appear. The border wrapped around the entire image+text paragraph until translating the caption, then the image only surrounded the image. After saving Media, Paragraph and Article, the content was indeed public even though admins see orange box around image.
  - [x] Photo Gallery — orange border reversed while editing.
  - [x] Achievements — orange border reversed while editing.
  - [x] Interactive :+1:
  - [x] Videos :+1:
  - [x] Section Index — confusing at first, but once the individual link destinations were translated then the Section Index filled itself out. The links appear to be in English even when the alternate language of that page was entered.. e.g. I entered https://gho.test/fr/ `Avant-propos du Coordonnateur des secours d'urgence (4)` but it still displays "Foreword by Emergency Relief Coordinator" while linking to http://gho.test/fr/introduction-and-foreword/avant-propos-du-coordonnateur-des-secours-durgence
- [ ] Noting red border for neither translated nor published — I never saw the red border. Everything that to me should have had it (e.g. an EN paragraph within FR Article) just had no border at all, as if it were ready to go.
- [ ] Noting dotted border around Media library items — never saw this and I didn't see a "Published" checkbox on the media itself so that makes sense.
- [x] Removing a Paragraph from a Translation so that it has fewer Paragraphs than EN.
- [x] Noted RTL layout with mostly EN UI when translating into Arabic

orakili:
@rupl Thanks for the detailed review. I've pushed several commits to fix the issue you found and some others.

First the changes, then some answers to your points + quirks will have to deal with/document.

## Changes

### Admin language

Previously, the language for the admin pages was forced to use English only. This caused several issues in the form, including what you experienced where the media in the a French form were flagged wrongly.

So instead, I went with a different approach and simply disable string translations on admin pages which is much less disruptive as it only means that UI texts will be in English and doesn't affect other things that rely on the current language.

To go with that, I removed the translations for the admin views (ax: /admin/content overviews) because Drupal handles that differently than the translations done via `t` and the like. This doesn't affect the functionality of the views, it just means that the UI texts for them are also always in English.

### Overviews

There is a new views field plugin that generates the the `edit/create` links for translations instead of using views inside views. This was done because the links were not generated properly due to various "views" reasons I don't feel like explaining.

The new approach is much simpler, will allow us to remove a module (`views_field_view`) and generate the prop links.

In addition, this adds a `view` to each existing translation for nodes so that it's much easier to view the translated nodes.

Screen Shot 2020-12-09 at 18 40 18

### Access checks

Access checks were added for paragraphs to properly hide them when not translated.

There is also some changes to add entity cache dependency where appropriate.

### Article list

The article list formatter was modified to properly handle translations, node accessibility and caching.

Also it adds the `--unpublished` and `--untranslated` classes on  the links so we can see what is going to be visible to the visitors.

## Answers

### Media translations

The above changes fix the issue with the inconsistency between the translation status of a media and what it shown in the Facts and figures, photo gallery and achievement paragraphs (what you showed in the video).

As you noticed, media are always published. This is on purpose because there is no need to add complexity here as they don't have a standalone page and the images are public anyway.

### Section index

The untranslated and/or unpublished links are now properly flagged and hidden to visitors and the proper translated titles are used for the links (when available).

### Red border flag

This only appears if you are on a translation and add a paragraph that references an unpublished node (sub-article, story or achievement) that doesn't have a translation either. So you were not seeing that because all English articles/stories/achievements are published.

To see it in effect, you could, for example, unpublish the video `story` on the foreword and then when creating the Spanish translation you should see it flagged in red.

## Quirks

### Nested paragraphs

When you create a translation of a node, you'll see that paragraphs inside paragraphs like the "Image and text" paragraphs of a "Facts and figures" paragraph will be flagged as untranslated. They will be so until the containing paragraph (the facts and figures) is saved again (even if there is no change).

It's not an issue per se but it can be a bit confusing for example for "Section index" paragraphs to see the article lists inside be flagged (they will not be displayed so it's consistent) while the `caption` field is not.

That's because Drupal creates a translation of the first level paragraphs when creating a node translation but doesn't create a translation of the nested paragraphs.

### Edit link

There are some cases where you (with administrator role) may have the impression to edit a translation while it's not the case. That happens when viewing a page in English, then switching to its non-existent translation and clicking edit.

For example: assuming https://gho.test/delivering-better/cash-and-voucher-assistance is not translated in French, you visit it in English, then switch to French with the language switcher and then click "edit", you end up editing the English version because the French translation doesn't exist yet and needs to be created via the "translate" tab or via the "create" link in the overview.

So we should strongly mention in the documentation to always check the page title to ensure it contains something like "Creating French translation of blabla" or "Editing Blablab [Francais translation]" to confirm one is editing/creating the proper translations.

I'll create a ticket for the above and see if I can find a way to twist Drupal into not being stupid.

html/modules/custom/gho_general/src/Plugin/LanguageNegotiation/GhoGeneralLanguageNegotiationAdmin.php
```diff
@@ -1,170 +0,0 @@
-<?php
-
-namespace Drupal\gho_general\Plugin\LanguageNegotiation;
```
rupl, line 3: I will spare you the full story, but I was unable to import config until I restored the language negotiation plugin. I believe this will be an issue when deploying to various environments.

The normal process of switching between `drush cim` and `drush cr` never seemed to break the cycle. The error I always saw for both `cim` and `cr` commands was:

```
  The "gho-general-language-admin" plugin does not exist. Valid plugin IDs for Drupal\language\Languag
  eNegotiationMethodManager are: language-browser, language-content-entity, language-selected, languag
  e-session, language-interface, language-url, language-url-fallback, language-user, language-user-admin
```

I'm skipping some steps in my investigation, but I eventually pulled DB dev snapshot and grepped it, finding three instances of `gho-general-language-admin` in the `config` table.

Through that error, I eventually found this deleted Plugin file in the diff and tried running the import with the file restored. I was able to import config and get the site running with the plugin in its old location.

I think the missing plugin combined with the DB entries is the problem. Those three entries will remain unless we run an update hook or other operation to clean them up before trying to import config or clear caches on the deployed environment.

rupl (changes requested):
I will continue QA in the mean time, but this initial issue seemed like a deploy blocker to me.

rupl:
Moving past the language negotiation:

### Overviews

Working well!

### Media

Media seems to be working much better now and I didn't need to translate nearly as many things to get images showing up on the page.

### Red Borders

I still never see a red border and I'm pretty much convinced they won't ever display if people are translating Articles containing fully-filled Paragraphs sections. Are they important? Do we need to be sure they're working?

If you do think it's likely and will cause them trouble or data loss, can you provide a video where you cause them to appear? Maybe Emma can make a judgement on whether the use-case will happen frequently.

The closest I can get is a yellow-dotted with red background. This is an unpublished text paragraph on a published Article translation:

image


### Section Index

This works very well and I think it makes a whole lot more sense, both in terms of data entry and when previewing the content

image


### Documentation Opportunities

**Translating Paragraphs which contain Entity Refs:** this is an issue with Paragraphs, but I think we need to very clearly mention that editing a Video, Story or Achievement from within an Article will cause you to lose any changes you had made beforehand. Using the dropdown and selecting Translate kicks you to a new page and all the unsaved work will be gone.

orakili:
@rupl I've added back the language negotiation plugin with a change so that it does nothing resulting in Drupal using the next plugin to determine the language. But that's just extra precaution because after the `drush cim` there will be no reference to that plugin anymore in the config table. I created GHO-183 to remove the plugin and some other unused modules after the current has been deployed to prod.

Created also GHO-184 to gather things that we should put in the documentation.

rupl (approved):
Another round of testing and no errors or anything major to speak of. Let's let the others try it on dev!

PR #187, opened 2020-12-03
[GHO-178] Make menu respect visibility of links

orakili:
Ticket: GHO-178

**Note**: this is a follow-up of #186 that should be merged first. Only the https://github.com/UN-OCHA/gho-site/pull/187/commits/2ebdfd697bfe032f5d72af964949990cfd00a58c commit here is relevant to GHO-178.

This replaces the block for the main navigation with a block that respects the language visibility of the content: i.e. only menu items for which there is a published page in the current language is displayed.

Note: the "Download report" is always displayed even if not translated. This is an opinionated choice (that would complicate the logic and it's a link to a page with all the versions so...).

## Testing

`drush cim`, `drush cr`, add some translations of the main articles and check they appear in the menu for the language.

## Screenshots

**Spanish with nothing translated**

Screen Shot 2020-12-03 at 23 59 31

**French with the foreword translated** (well actually text is still in English but the translation exists and is published)

Screen Shot 2020-12-04 at 0 00 32


orakili:
Replaced with https://github.com/UN-OCHA/gho-site/pull/188 which is independent from https://github.com/UN-OCHA/gho-site/pull/186

PR #188, opened 2020-12-07
[GHO-178] Main menu

orakili:
Ticket: GHO-178

This replaces the block for the main navigation with a block that respects the language visibility of the content: i.e. only menu items for which there is a published page in the current language is displayed.

Note: the "Download report" is always displayed even if not translated. This is an opinionated choice (that would complicate the logic and it's a link to a page with all the versions so...).

## Testing

`drush cim`, `drush cr`, add some translations of the main articles and check they appear in the menu for the language.

## Screenshots

**Spanish with nothing translated**

Screen Shot 2020-12-03 at 23 59 31

**French with the foreword translated** (well actually text is still in English but the translation exists and is published)

Screen Shot 2020-12-04 at 0 00 32


html/themes/custom/common_design_subtheme/templates/block/block--gho-main-menu.html.twig
```diff
@@ -0,0 +1,59 @@
...
+    'block',
+    'block-menu',
+    'navigation',
+    'menu--' ~ derivative_plugin_id|clean_class,
+    'cd-nav',
+    'cd-site-header__nav',
+  ]
+%}
+
+{% set heading_id = attributes.id ~ '-menu'|clean_id %}
```
rupl, line 45: as we've seen in other cases, this does not sanitize the dynamic part of the HTML ID being constructed, only `'-menu'` gets sanitized by `clean_id` filter.

```suggestion
{% set heading_id = (attributes.id ~ '-menu')|clean_id %}
```

rupl (approved):
Works like a charm. Noted one minor nit, and was even able to format a suggestion this time so  you can click it into place 👍

rupl:
Rather than have this sit out there and need three updates to dev, I committed my change and will merge shortly.

PR #189, opened 2020-12-07
[GHO-180] Drupal core and scaffold update

orakili:
Ticket: GHO-180

This changes the drupal scaffold module, updates core to 8.9.1 and add a post install/update command to clean after symfony/flex.

**Notes**

- Had to let the `.editorconfig` be replaced because I couldn't get drupal-scaffold to preserve the one we had and it's not a big deal
- Added a `removeUnnecessaryFiles` function the `ScriptHandler` to clean after symfony/flex
- Changed the `file-mapping` for the `settings.php` and `services.yml` to preserve existing files but for the `settings.php` it's probably unnecessary as the `ScriptHandler` does basically the same.


.editorconfig
```diff
@@ -13,8 +13,5 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[composer*.{json,lock}]
+[composer.{json,lock}]
 indent_size = 4
```
cafuego, line 17: This hurts.

cafuego (approved):
That looks OK and seems to work as advertised.

.editorconfig
orakili: I guess drupal doesn't want us to format the patch file or the makefile cleanly...

rupl:
I did my classic thing and screwed up my local but no sense in holding this up if my stuff is broken either way.

PR #191, opened 2020-12-09
Set PHP to 7.3 in composer.json

rupl:
Recommended by @lazysoundsystem for my PHP woes after updating composer (output by Drupal as best I can tell)

```
Fatal Error: composer.lock was created for PHP version 7.4 or higher but the current PHP version is 7.3.18.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,6 +75,9 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "config": {
+        "platform": {
+            "php": "7.3"
+        },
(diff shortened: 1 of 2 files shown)


rupl:
I forgot to test this. Now it includes the composer.lock which Composer knew they weren't in sync (I think via `platform-overrides` not existing)

lazysoundsystem (approved):
Looks right to me. 

(Only issue with this is that we have to remember to update it when we get php7.4 containers, but having it on multiple sites will make it easier to remember.)

PR #192, opened 2020-12-09
[GHO-134] create option to use native lazy loading

rupl:
# GHO-134

This is a draft because it's not working yet, but I'm a little bit stuck in terms of how to proceed. I thought maybe you will know what the problem is and at least I got this far on my own.

The "Responsive image" formatter now has a new dropdown to select what value the `` attribute will have. If an image lacks the attribute, it's treated as `auto` (which results in historical prefetching behavior).

I went through a couple iterations of this and settled on the method here, which @attiks recommended. It uses third-party settings hooks for the formatter. More info here: https://www.drupal.org/node/2130757

Anyway, I looked at this until my eyes crossed and I feel like it is fairly close, but I never see the attribute on an image using the intended method of choosing the option in the dropdown of the formatter. I can hardcode the `preprocess_image` to display the attribute, so I know if the attribute were there, the standard image template would render it.

When I dump `array_keys($variables)` inside `common_design_subtheme_preprocess_image()` I get this list:

```
Array
(
    [0] => uri
    [1] => alt
    [2] => attributes
    [3] => width
    [4] => height
    [5] => title
    [6] => sizes
    [7] => srcset
    [8] => style_name
    [9] => theme_hook_original
    [10] => title_attributes
    [11] => content_attributes
    [12] => title_prefix
    [13] => title_suffix
    [14] => db_is_active
    [15] => is_admin
    [16] => logged_in
    [17] => user
    [18] => directory
    [19] => #cache
    [20] => is_front
    [21] => username
    [22] => componentsModule
    [23] => base_path
)
```

I don't see `element` there at all. That makes me think I'm not even close.

Is this a paragraphs thing, and actually several layers of recursion removed from where I really want to inject my attribute? I thought _maybe_ I wouldn't have to factor Paragraphs into this, since somewhere down the line there's a field and I am trying to preprocess it.

## End Goal

If the View Mode's `responsive_image` formatter has `loading: lazy` selected, then the image tag should render as:

```html
<img ... loading="lazy" lazyload="1">
```


orakili:
@rupl Here you go: https://github.com/UN-OCHA/gho-site/pull/193.

I think we would need to add a few **media view modes** to make sure the lazy loading is not added (or set to auto) to the hero image of articles in `full` view mode and for the image of stories in `full` mode as well. Something like:

- Hero image - Full
- Story - Full

Also I think it would be good to add something in the css components to avoid layout shifts (we're lucky we have fixed aspect ratios): https://css-tricks.com/aspect-ratio-boxes/#the-psuedo-element-tactic

rupl:
naming View Modes seems to be the hardest problem in Drupal 😁 

Now that it's actually working, I'll play around and see which ones are totally safe (Related Articles) and which need some adjusting. Once they are configured I'll ping for a functional review.

orakili (approved):
That works well. Also the hero images of the section index paragraphs have the lazy loading attribute so forget what I said during the standup, I just overlooked it.

PR #193, opened 2020-12-10
[GHO-134] Add lazy loading attributes to images

orakili:
Ticket: GHO-134

Follow-up of #192 

Screen Shot 2020-12-10 at 13 45 40


rupl:
Well at least I was mostly on the right track. Thanks so much!

rupl:
Now that it's all in one module, that makes me think we could keep this around as a general perf improvement on many sites...

PR #194, opened 2020-12-11
[GHO-168] Search

orakili:
Ticket: GHO-168

This enables the drupal core search module, adds the search block in the header and a search results page with basic styling.

Screen Shot 2020-12-11 at 17 13 35

## Issue

There is a real estate problem in the header and at in English, at around 1100px, the navigation header starts not respecting the right padding:

Screen Shot 2020-12-11 at 17 49 49

It's worse in French for example where the navigation button labels are longer.

One solution would be to break before 1024px and so show the "Menu" button earlier for the header but it's not easy to tweak so it's not included in that PR.

## Testing

`drush cim`, `drush cr` and search for something using the header search button.


orakili:
@rupl, sorry, I simply forgot to push the changes with the permissions. I've just added a commit for that to allow anonymous and authenticated users to use the simple search (no advanced to keep things simple for now).

emmajane:
Thanks for double checking permissions. It would be good to get this on everyone's personal checklist: test everything on anon AND user with "editor" role for the site.

rupl:
Closing in favor of #245

PR #198, opened 2020-12-15
[GHO-185] Ensure the homepage content is visible to editors

orakili:
Ticket: GHO-185

This removes the "is promoted" condition to display the homepage's content that was added a while ago when there was still the idea of showing the homepage hero image and tagline for the Arabic, French and Spanish versions while waiting for the content to be translated.

This leftover is currently preventing the preview/view of a translation of the homepage when the "global visibility" checkbox was not checked.

The current workaround is to have the "global visibility" checkbox checked and the "published" checkbox unchecked, which preserves the behavior of hiding content from the visitors until both are checked while allowing editors to preview/view the homepage's content in a given language.

So this "fix" is simply to allow the double security of having both the "global visibility" and the "published" unchecked while content is being translated.


rupl (approved):
Works like a charm :shipit:

PR #199, opened 2020-12-15
[GHO-29] Article PDF download

orakili:
Ticket: GHO-29

This is the first part of GHO-29 to create upload PDFs to articles and have a download URL. Next PR will integrate that with the work on GHO-7.

This provides a route to have permanent download URLs for the PDF version of an article: https://gho.unocha.org/node/NODEID/download.

Drupal being what it is, the original filename is not saved... so unfortunately the filename when saving the above url may have the `_X` suffix. [0]

It's also not possible to have drupal rename the file to preserve its URL. [1]

- [0] https://www.drupal.org/project/drupal/issues/3032376
- [1] https://www.drupal.org/project/drupal/issues/2648816

This is inspired from https://www.drupal.org/project/media_entity_download

## Testing

`drush cim`, `drush cr` then add a file to an article and access it via `/node/NODEID/download`.

### DIFF
--- a/config/core.entity_form_display.node.article.default.yml
+++ b/config/core.entity_form_display.node.article.default.yml
@@ -8,6 +8,7 @@ dependencies:
     - field.field.node.article.field_caption
     - field.field.node.article.field_hero_image
     - field.field.node.article.field_paragraphs
+    - field.field.node.article.field_pdf
     - field.field.node.article.field_report_link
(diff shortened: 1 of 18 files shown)


rupl (approved):
Work great! Tested anonymously in Chrome/FF/IE11 and it worked everywhere when I directly visited a URL such as https://gho.test/node/47/download

PR #200, opened 2020-12-15
[GHO-29] Display download icon when there is a PDF version of an article

orakili:
Ticket: GHO-29

See PR's title.


rupl (approved):
Worked well in macOS Chrome/FF and Win7 IE11. Since it's all twig I didn't expect otherwise, but all the browsers load the PDF so the content-disposition header does seem to be working everywhere.

PR #202, opened 2020-12-15
[GHO-179] Official large logos

orakili:
Ticket: GHO-179

Official large logos for the homepage.


rupl (approved):
the logos are indeed updated :shipit:

PR #204, opened 2020-12-16
[GHO-7][GHO-173] Social icons and appeal tags fixes

orakili:
Tickets: GHO-7 and GHO-173

This PR adds some styling fixes to avoid overlaps, have more consistent alignments for the social icons and appeal tags.

### Social icons issues

- Social icons overlap with pre-title or title on small screens

Screen Shot 2020-12-16 at 17 04 44
Screen Shot 2020-12-16 at 17 06 02

- Social icons alignment is not consistent across browsers (looking at you Safari)

Screen Shot 2020-12-16 at 17 08 28

- According to prototype document, the social icons are supposed to be under the titles on small screens (page 20):

Screen Shot 2020-12-16 at 17 10 11
Screen Shot 2020-12-16 at 17 10 17

- The `ul` of the social icons is over the `pre-title` for example preventing text selection.

### Appeal tags issues

- Appeal tags don't appear on small screens when title is long

Screen Shot 2020-12-16 at 17 06 43

- Appeal tags overlap titles

Screen Shot 2020-12-16 at 17 12 00

### Social icons

This uses a different approach for the social icons, using a 50% padding-top to place them as it's based on the width of the container (the article tag) on medium and large screens and place them between the title and the content on small screens as per the prototype use the `order` property.

The `ul` was put in a wrapper so that we could use a `position: absolute` on the `ul` in order to limit its width to its content and so not overlap with the pre-title or title which was preventing text selection.

Screen Shot 2020-12-16 at 17 16 07
Screen Shot 2020-12-16 at 17 16 21

### Appeal tags

This changes the display of the article/sub-article title to `inline` so that the appeal tags stay attached to the title even when it's long. This required to remove the reading width `max-width` on the title, so to preserve this width, a variant of the `content-width` class is introduced which limits the content to the `reading-width` while the container still behaves as having the `content-width` class.

Screen Shot 2020-12-16 at 17 23 48

This also adjusts a bit the styling and move the rules to the `gho-appeal-tags` component. Bonus: the changes result in the text being properly vertically centered inside the colored rectangular as they wished for.

### Story and Achievement pages

Minor changes to ensure they use the `content-width` as the other pages + story title styling.

## Testing

`drush cr`


rupl:
I only found one issue which happens on mobile viewport and it's adjustable. I'm happy to pick this up:

image

html/themes/custom/common_design_subtheme/components/gho-social-links/gho-social-links.css
```diff
@@ -1,20 +1,47 @@
...
   justify-content: flex-end;
-  margin: 0 -4px;
+  margin: 0;
   padding: 0;
 }
+html[dir='ltr'] .gho-social-links ul  {
+  justify-content: flex-start;
+}
+html[dir='rtl'] .gho-social-links ul  {
+  justify-content: flex-end;
```
rupl, line 20: browsers do this for us when detecting RTL. I'll remove when I commit the other tweaks to "Copy URL" styles

orakili:
Thanks, I didn't think of testing the copy URL message.

rupl:
@orakili if you could re-review your own PR I've popped some commits on there including:

- RTL layout fix for that `flex-end`
- The feedback bubble
- While testing I noticed FF showed serif latin text on Arabic pages. I adjusted the global typography rule for Arabic pages to fix in FF.

### Before in FF

image

### After in FF

image

html/themes/custom/common_design_subtheme/components/gho-social-links/gho-social-links.css
orakili: Ok, cool, thanks. I got things mixed up at some point when testing RTL and that fixed the issue but this was before using the `absolute` positioning. It looks like it's not even necessary anymore in the base `.gho-social-links ul  {...}` with the absolute position as the `ul` has the width of its content.

orakili:
@rupl Reviewed and tested. Thanks!

rupl:
I feel like it's good now. Here we go!

PR #205, opened 2020-12-16
[GHO-186] translation files

rupl:
~~This PR has absolutely no effect on the website. I am simply updating the PO exports based on the latest rounds of data entry.~~ Whoops sorry that's not true. I did include some core administrative config translation updates.

I guess `drush cim` and `drush cr` and then import the POs using procedure in `translations/README.md`


rupl:
I should add that although the diffs for POs are large and things look like they're being deleted, I checked them and they all get added back in at a different location within the files.

orakili (commented):
Drupal is annoying, I explicitly deleted the views translations so that the admin interface stays in English instead of being partially translated and it's adding them back...

I think it does that when installing new modules or making changes to content types...

Well, I can live with it. Just another Drupal's 💩 

## Arabic issues

- The "OCHA services" translation is way too long and makes a mess in the top header on mobile:

Screen Shot 2020-12-16 at 19 34 53

- The CERF and CBPF are not linked
 
Screen Shot 2020-12-16 at 19 36 44

emmajane:
This is fine to leave for a deployment in 2021.

rupl:
Yes I believe it happened when I re-enabled stage_file_proxy.. shoot.

PR #207, opened 2021-01-04
[GHO-183] Module update and cleanup

orakili:
Ticket: GHO-183

This removes the `views_field_view` module which was used for the admin content overviews but was replaced with a custom plugin and the language negotiation plugins that was replaced with a string translation plugin to have the admin UI in English.

This also contains some module updates from running `composer update`, the main one being the phpspreadsheet library. Needs and requirements import was tested with this new library and it still working as expected.

## Testing

`composer update` then `drush cr`.


orakili:
@rupl bump.

rupl (approved):
I am approving the opportunity to view this on dev.

PR #208, opened 2021-02-01
[GHO-188] update translations

rupl:
# GHO-188

No effect on website, but updating the individual PO files with latest translations, along with a couple tweaks to templates to support them.


orakili (approved):
Thanks.

html/themes/custom/common_design_subtheme/templates/nodes/node--story--full.html.twig
```diff
@@ -100,7 +100,7 @@
 <article{{ attributes.addClass(classes) }}>
   <header class="gho-story__header content-width">
     {% if pre_title %}
-    <span class="gho-aside__pre-title">{{ pre_title }}</span>
+    <span class="gho-aside__pre-title">{{ pre_title|t }}</span>
```
rupl, line 103: I'm now realizing that these aren't translated in the theme layer because they are translated as content within the Taxonomy system. I guess I can back this change out at some point.

html/themes/custom/common_design_subtheme/templates/taxonomy/taxonomy-term--appeals.html.twig
```diff
@@ -30,7 +30,8 @@
     'gho-appeals-tag',
   ]
 %}
-{% set tagname = name|render|striptags|trim %}
-<div{{ attributes.addClass(classes, ('gho-appeals-tag--' ~ tagname)|clean_class) }} title="{{ content.field_appeal_extended|render|striptags|trim }}">
+{% set tagname = name|render|striptags|trim|t %}
+{% set extended = content.field_appeal_extended|render|striptags|trim|t %}
```
rupl, line 34: same here with the addition of `|t` to both values. It wasn't necessary and I was trying to translate at theme/UI layer instead of directly translating taxonomy terms.

PR #217, opened 2021-02-03
[GHO-192] Use new OCHA services template

orakili:
Ticket: GHO-192

This moves the OCHA services template to the footer and use `block` twig comments to override the links instead of the entire template for easier maintenance.

## Testing

`drush cr` and check that the correct links are displayed instead of `customizable link`.


left23 (commented):
@orakili There might be some colour adjustments needed for when JS is disabled, due to the inverted footer colours.
Screenshot from 2021-02-03 11-29-06

rupl:
I will handle the no-JS styling

rupl (approved):

PR #218, opened 2021-02-03
[GHO-188] update translations with prod exports

rupl:
# GHO-188

Not urgent!

I did a bunch of data entry on prod and wanted to store it in version control.


orakili (approved):
Thanks.

PR #233, opened 2021-04-30
Gho 196 poc

attiks:


berliner (approved):
Looks good to, without being actually able to judge this in any way ;)

cafuego:
I concur with Travis the PR should install the workspaces module. 🙅🏼‍♂️

cafuego (changes requested):
Workspaces is missing from core.extension.yml

PR #248, opened 2021-08-17
[OPS-7688] core security update

lazysoundsystem:
https://humanitarian.atlassian.net/browse/OPS-7688


lazysoundsystem:
@rupl just checking that the styling changes already in dev are all okay to go.

config/block.block.searchform.yml
```diff
@@ -0,0 +1,21 @@
+uuid: 3ab832bf-8def-4c08-93e7-da8803515718
+langcode: en
+status: true
```
rupl, line 3: @lazysoundsystem the only issue is that I enabled search via config and until we get sign-off on how dev looks, I don't think it's appropriate to deploy.

@emmajane do you agree that we should disable search before deploying this work to prod?

rupl: (I picked this line because I think changing `true` to `false` is enough to hide the block and avoid exposing search to the public)

emmajane: tl;dr: it should not display. If it's enabled but not showing up I guess that's okay if it's easier to get the security update out. But it should not display. (I said that bit twice, right? :) )

We're still a few weeks away from wanting to enable search on prod. I have a small preference for keeping it disabled on prod until we're ready for it. I have a large preference for it not displaying in the main menu until it's been approved by the team.

rupl (changes requested):
Based on @emmajane comments I am requesting changes and assigning myself this PR until `develop` is updated with necessary config to do the following:

- Disable search entirely.
- Make sure the Block that appears in the main nav is not showing up.

rupl:
With the latest commits to `develop` I believe this is ready.

rupl (approved):

PR #249, opened 2021-08-18
OPS-7688: disable Search which wasn't prod-ready

rupl:
# OPS-7688

Unblocks the security update (I think).

## Testing

Import config and make sure the main nav no longer has a search widget.


lazysoundsystem (approved):
Looks right to me - this leaves in 'develop' some search-related css that's not being used, yet, but I don't think there's any harm in that.

rupl:
yup, that's intentional. When we're ready to use search, we'll be able to turn it back on without additional adjustments to the theme.

PR #256, opened 2021-10-13
[OPS-7743] add and configure seckit in report mode

lazysoundsystem:
https://humanitarian.atlassian.net/browse/OPS-7743

Adds and configures seckit module in report mode.

### DIFF
--- a/config/seckit.settings.yml
+++ b/config/seckit.settings.yml
@@ -1,0 +1,56 @@
+seckit_xss:
+  csp:
+    checkbox: true
+    vendor-prefix:
+      x: false
+      webkit: false
+    report-only: true
+    default-src: ''
+    script-src: '''self'' ''unsafe-inline'' https://fonts.googleapis.com https://www.gstatic.com https://www.google.com https://www.google-analytics.com https://cdn.jsdelivr.net'
+    object-src: ''
+    img-src: '''self'' data:'
+    media-src: ''
+    frame-src: ''
+    frame-ancestors: ''
+    child-src: ''
+    font-src: 'https://fonts.gstatic.com'
+    connect-src: '''self'' https://www.google-analytics.com https://stats.g.doubleclick.net'
+    report-uri: /report-csp-violation
+    upgrade-req: false
+    policy-uri: ''
+    style-src: ''
+  x_xss:
+    seckit_x_xss_option_disable: Disabled
+    seckit_x_xss_option_0: '0'
+    seckit_x_xss_option_1: 1;
+    seckit_x_xss_option_1_block: '1; mode=block'
+    select: 0
+seckit_csrf:
+  origin: false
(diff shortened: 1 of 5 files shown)


lazysoundsystem:
@rupl not much to check here as we're going to run seckit on the live site in 'report' mode before switching it on properly, so we can pick up anything that might need to be added to the config. But might as well be deployed at the same time as https://github.com/UN-OCHA/gho-site/pull/257 .

(I'll fix the core.extension.yml conflict from the other PR now)

config/seckit.settings.yml
```diff
@@ -0,0 +1,56 @@
...
+    media-src: ''
+    frame-src: ''
+    frame-ancestors: ''
+    child-src: ''
+    font-src: 'https://fonts.gstatic.com'
+    connect-src: '''self'' https://www.google-analytics.com https://stats.g.doubleclick.net'
+    report-uri: /report-csp-violation
+    upgrade-req: false
+    policy-uri: ''
+    style-src: ''
```
rupl, line 21:
```suggestion
    style-src: '''self'' https://fonts.googleapis.com'
```

I think we need the following modification to display Google Fonts. I used HID CSP as my reference.

config/seckit.settings.yml
```diff
@@ -0,0 +1,56 @@
+seckit_xss:
+  csp:
+    checkbox: true
+    vendor-prefix:
+      x: false
+      webkit: false
+    report-only: true
+    default-src: ''
+    script-src: '''self'' ''unsafe-inline'' https://fonts.googleapis.com https://www.gstatic.com https://www.google.com https://www.google-analytics.com https://cdn.jsdelivr.net'
```
rupl, line 9:
```suggestion
    script-src: '''self'' ''unsafe-inline'' https://fonts.googleapis.com https://www.gstatic.com https://www.google.com https://www.google-analytics.com https://www.googletagmanager.com https://cdn.jsdelivr.net'
```

```
[Report Only] Refused to load the script 'https://www.googletagmanager.com/gtm.js?id=GTM-xxxxxx' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://www.gstatic.com https://www.google.com https://www.google-analytics.com https://cdn.jsdelivr.net". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

rupl (approved):
Since it's report-only, we can perhaps wait to make the modifications I suggested. However GTM warning did show up on my local. The fonts are just my suspicion but we do have this markup in the HTML:

```html
<link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;0,900;1,400;1,700&display=swap" rel="stylesheet">
```

PR #260, opened 2021-11-11
GHO-225: Article Card, Article Card List, Heading paragraph types

rupl:
# GHO-225

We want the Article Card style for the section indices on 2021 site too.

## Testing

As this is an identical copy of https://github.com/UN-OCHA/gho-2022-site/pull/12 the testing instructions are there, save for pulling the snapshot from the 2021 dir, not "dev" which holds 2022.

### DIFF
--- a/.docksal/docksal.env
+++ b/.docksal/docksal.env
@@ -14,7 +14,7 @@ DB_IMAGE="docksal/db:1.1-mysql-5.7"
 CLI_IMAGE='docksal/cli:php8.0-build'
 
 # Override virtual host (matches project folder name by default)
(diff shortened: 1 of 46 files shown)


lazysoundsystem (approved):
Looks good. And tested the cards are clickable too.

Minor thing I noticed is that the card contents are slightly offset against the card outline, so the border on focus doesn't quite line up. I only had a quick look, but I couldn't see why not.

image

rupl:
Good spot! I _believe_ this won't matter because all the content we'll be using cards for have Hero images. I quickly clicked through every entry on prod just now.

PR #262, opened 2021-11-12
GHO-226: Article Card List summaries

rupl:
# GHO-226

We're trying to enforce consistent styling of the Card Lists across websites, and the 2022 site uses a limit of 120 characters. When I sat down to implement this, I thought it would be a matter of changing a number in the appropriate View Mode, but I ran into issues with the behavior. Specifically, under the trim setting it says:

> If the summary is not set, the trimmed Summary field will end at the last full sentence before this character limit.

In practice I found this to mean the summaries had wildly different lengths based on whether a sentence ended within that 120-char window. My solution was to "un-trim" the summary coming out of the View Mode and do it directly in the Twig template.

So you get pretty darn consistent summary sizes, but it also is aware enough to skip the treatment for short summaries. See screenshot:

Screen Shot 2021-11-12 at 15 18 44

### DIFF
--- a/config/core.entity_view_display.node.article.teaser_card.yml
+++ b/config/core.entity_view_display.node.article.teaser_card.yml
@@ -48,12 +48,11 @@ content:
     type: entity_reference_label
     region: content
   field_summary:
-    type: text_trimmed
+    type: text_default
     weight: 2
     region: content
     label: hidden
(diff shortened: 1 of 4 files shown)


attiks (approved):
Twig++

